PR #31, opened 2020-04-21
Improve error types to allow wrapping

### LINKED ISSUE #13: piv: better errors for apdu error codes
https://www.eftlab.com/knowledge-base/complete-list-of-apdu-responses/

Common ones:

sw1=0x6a, sw2=0x80 - incorrect input or unsupported algorithm
sw1=0x69, sw2=0x82 - unauthenticated
sw1=0x69, sw2=0x83 - authentication method blocked

dnesting:
I got stymied not having an idiomatic way of seeing whether a smartcard object was not found or not, and I saw the TODO in the code to list error cases, so I thought I'd offer this improvement for comment.

- Promotes `apduError` to a public `ApduError` type
- Adds error messages from the latest PIV spec
- Uses error `Unwrap()` to return a few more accessible/idiomatic Go errors to enable some `errors.Is` and `errors.As` uses, such as `ErrNotFound`, `ErrBlocked`
- Promotes `errWrongPin` to `VerifyErr` type
- Changes all `%v` verbs to `%w` when wrapping errors (new in 1.13, which is what `go.mod` says we are)
- The above changes made `ykTransmit` unnecessary, so removed

I can now do things like:

```go
// Don't overwrite a slot
_, err = yk.Certificate(slot)
if err == nil {
  return errSlotOccupied
} else if !errors.Is(err, piv.ErrNotFound) {
  return err
}
```

Fixes #13

### DIFF
--- a/piv/key.go
+++ b/piv/key.go
@@ -319,9 +319,9 @@ func ykAttest(tx *scTx, slot Slot) (*x509.Certificate, error) {
 		instruction: insAttest,
 		param1:      byte(slot.Key),
 	}
-	resp, err := ykTransmit(tx, cmd)
+	resp, err := tx.Transmit(cmd)
 	if err != nil {
-		return nil, fmt.Errorf("command failed: %v", err)
+		return nil, fmt.Errorf("command failed: %w", err)
 	}
 	if bytes.HasPrefix(resp, []byte{0x70}) {
 		b, _, err := unmarshalASN1(resp, 0, 0x10) // tag 0x70
@@ -360,9 +360,9 @@ func ykGetCertificate(tx *scTx, slot Slot) (*x509.Certificate, error) {
 			byte(slot.Object),
 		},
 	}
-	resp, err := ykTransmit(tx, cmd)
+	resp, err := tx.Transmit(cmd)
 	if err != nil {
-		return nil, fmt.Errorf("command failed: %v", err)
+		return nil, fmt.Errorf("command failed: %w", err)
 	}
 	// https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-73-4.pdf#page=85
 	obj, _, err := unmarshalASN1(resp, 1, 0x13) // tag 0x53
@@ -407,7 +407,7 @@ func (yk *YubiKey) SetCertificate(key [24]byte, slot Slot, cert *x509.Certificat
 	}
 	defer tx.Close()
 	if err := ykAuthenticate(tx, key, yk.rand); err != nil {
-		return fmt.Errorf("authenticating with management key: %v", err)
+		return fmt.Errorf("authenticating with management key: %w", err)
 	}
 	return ykStoreCertificate(tx, slot, cert)
 }
@@ -433,7 +433,7 @@ func ykStoreCertificate(tx *scTx, slot Slot, cert *x509.Certificate) error {
 		param2:      0xff,
 		data:        data,
 	}
-	if _, err := ykTransmit(tx, cmd); err != nil {
+	if _, err := tx.Transmit(cmd); err != nil {
 		return fmt.Errorf("command failed: %v", err)
 	}
 	return nil
@@ -461,7 +461,7 @@ func (yk *YubiKey) GenerateKey(key [24]byte, slot Slot, opts Key) (crypto.Public
 	}
 	defer tx.Close()
 	if err := ykAuthenticate(tx, key, yk.rand); err != nil {
-		return nil, fmt.Errorf("authenticating with management key: %v", err)
+		return nil, fmt.Errorf("authenticating with management key: %w", err)
 	}
 	return ykGenerateKey(tx, slot, opts)
 }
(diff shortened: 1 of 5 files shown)


ericchiang:
I'm fine with ErrNotFound and improving the error message. Please add a test for that one :)

Less convinced that we need to expose so much API surface to deal with errors. Maybe ErrWrongPIN is worth exposing, but what practical program is going to do anything except return that error to the user? Definitely think that we shouldn't expose apduError. It way too low level, and I'd rather add specific error conditions to the API, like an object not being found.

piv/key.go
```diff
@@ -98,7 +98,7 @@ func (a *Attestation) addExt(e pkix.Extension) error {
 	} else if e.Id.Equal(extIDSerialNumber) {
 		var serial int64
 		if _, err := asn1.Unmarshal(e.Value, &serial); err != nil {
-			return fmt.Errorf("parsing serial number: %v", err)
+			return fmt.Errorf("parsing serial number: %w", err)
```
ericchiang, line 101: Please only wrap errors that are explicitly need to be passed through. I definitely don't want to support users errors.As'ing to things like asn1 error types :)

This applies to a lot of the s/%v/%w/g changes. Please revert those.

piv/key.go
```diff
@@ -319,20 +319,20 @@ func ykAttest(tx *scTx, slot Slot) (*x509.Certificate, error) {
 		instruction: insAttest,
 		param1:      byte(slot.Key),
 	}
-	resp, err := ykTransmit(tx, cmd)
+	resp, err := tx.Transmit(cmd)
```
ericchiang, line 322: Why was this changed? Mind reverting the diff?

piv/pcsc.go
```diff
@@ -48,16 +49,91 @@ func (e *scErr) Error() string {
...
+
+// ErrBlocked is returned when the authentication method on the smart card is blocked,
+// such as when you exhaust PIN retries.
+var ErrBlocked = errors.New("authentication method blocked")
+
+// ApduErr is an error interacting with the PIV application on the smart card.
+// This error may wrap more accessible errors, like ErrNotFound, ErrBlocked, or
+// an instance of VerifyErr, so callers are encouraged to use errors.Is and errors.As
+// for these common cases.
+type ApduErr struct {
```
ericchiang, line 75: Commented but I don't think this should be exported, though I'm fine with improving error messages for Error() and it supporting Unwrap().

piv/piv.go
```diff
@@ -347,13 +301,13 @@ func ykReset(tx *scTx, r io.Reader) error {
 			// TODO: do we care about a 1/100million chance?
 			return fmt.Errorf("expected error with random pin")
 		}
-		var e *errWrongPIN
-		if !errors.As(err, &e) {
-			return fmt.Errorf("blocking pin: %v", err)
-		}
-		if e.Retries == 0 {
+		if errors.Is(err, ErrBlocked) {
```
ericchiang, line 304: Think this block shows why we should just have one error type. Maybe s/VerifyErr/AuthErr/g (Verify sounds like it's an issue with a signature).

```go
var e *AuthErr
if !errors.As(err, &e) {
	return fmt.Errorf("blocking pin: %v", err)
}
if e.Retries == 0 {
	break
}
```

dnesting:
OK, that makes sense.  I'll pull this back to expose less, which means the `%w` changes add less value so I'll drop those as well.

The `ykTransmit` drop was intentional, though, as this function was essentially a no-op after the error handling got moved out of it.  This will be clearer with the reversions suggested.

piv/key.go
dnesting: Done.   I only kept those that were likely to wrap an `apduErr`.

piv/key.go
dnesting: With the addition of errors wrapped by `apduErr`, the logic in `ykTransmit` can be removed.  This leaves `ykTransmit` essentially empty except for the call to `tx.Transmit`.  Would it be helpful to see that reverted?

piv/pcsc.go
dnesting: I returned this to its unexported form.  However, callers will not be able to access the PIV error codes without parsing it out of the string message.

piv/key.go
ericchiang: Woops yeah I see that. lgtm.

piv/pcsc.go
ericchiang: Thanks. I think that's good for now. We can change it in the future if we want :)

piv/piv.go
dnesting: So I was operating under the belief that you would want to distinguish between the retries==0 case and the blocked case, but I see from experimentation that it goes from retries=2 retries=1 to blocked, so representing blocked as AuthErr(retries=0) seems safe?  I'll make that change.

dnesting:
OK I believe I've made all of the changes you suggested and added tests.  PTAL.

piv/pcsc.go
```diff
@@ -48,16 +49,92 @@ func (e *scErr) Error() string {
 	return fmt.Sprintf("unknown pcsc return code 0x%08x", e.rc)
 }
 
+// AuthErr is an error indicating an authentication error occurred (wrong PIN or blocked).
+type AuthErr int
```
ericchiang, line 53: nit: can this stay as a struct? don't know if there might be additional data we might want to add, but just to play it safe.

```go
type AuthErr struct {
    Retries int
}
```

dnesting: Sure.

dnesting:
Changes made and squashed.

I also made one additional small change to the error message text returned by `apduErr.Error()` for the 'blocked' case, to allow someone to distinguish between "verification failed" and "method blocked" there since `AuthErr` has the same value for each.  And added a test for that.

ericchiang:
Thanks for the couple iterations :) If you hit any other use cases where it'd be good to expose another error type, please let me know. Will tag a release later this evening

PR #32, opened 2020-04-22
Fix panic when k.begin() attempts to return early

dnesting:
https://github.com/go-piv/piv-go/blob/master/piv/key.go#L537-L541

The `defer` calls `tx.Close()` when needed to clean up, however all of the early return paths explicitly return `nil` for `tx`, causing a nil dereference panic.  This PR has two commits, one with the test case proving the panic and the other with a fix.

Another approach could be setting `err = fmt.Errorf(...)` and just doing a bare return at the end.


ericchiang:
Thanks!

PR #37, opened 2020-04-25
piv: only call PINPrompt when required

### LINKED ISSUE #35: PINPrompt should only be called if the PIN is needed
This logic would likely be:
* Try to sign
* On an error, request the PIN

ericchiang:
Fixes #35

fyi @FiloSottile

For some reason this doesn't work:

```
% go test -v -run=TestPINPrompt . -wipe-yubikey
=== RUN   TestPINPrompt
    TestPINPrompt: key_test.go:124: expected PINPrompt to only be called once got: 2
--- FAIL: TestPINPrompt (1.96s)
FAIL
FAIL    github.com/go-piv/piv-go/piv    2.253s
FAIL
```

May PINPolicyOnce still only applies to the smartcard transaction? Will investigate.

### DIFF
--- a/piv/key.go
+++ b/piv/key.go
@@ -24,6 +24,7 @@ import (
 	"crypto/x509/pkix"
 	"encoding/asn1"
 	"encoding/pem"
+	"errors"
 	"fmt"
 	"io"
 	"math/big"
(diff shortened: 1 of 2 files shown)


ericchiang:
Okay yeah, the PINPolicy only applies to the smartcard transaction. Which implies that we'd have to hold a transaction to the smartcard for this to work.

Maybe we can refactor the YubiKey struct to hold a transaction instead just a connection? It'd mean you couldn't have multiple open... but 🤷

Issue #47, opened 2020-04-26
See if cached PIN signing without requiring a PS/SC transaction


ericchiang:
Based off my testing, releasing the transaction wipes the PIN.

FiloSottile:
Alternatively, the same high-level use case would be served by having a "weak" lock on the token, by somehow letting other instances take over. This could be a PCSC option, a takeover method, a notification of other pending transactions, or even just a way to find out what process has an open transaction, letting the applications sort it out with signals.

ericchiang:
#51 now marks the smartcard context as exclusive so other applications that try to access the card will fail instead of hang.

Did a little looking into SCardGetStatusChange to see if we can detect when another application attempts to access the card. That might be a way to implement a "weak" lock.

https://ludovicrousseau.blogspot.com/2018/02/how-to-use-scardgetstatuschange.html

tv42:
Would it be possible to release the USB device, as a whole?

I have a Chromebook with multiple LXD containers ("crostini") where I would like to be able to SSH out from different containers using the same yubikey. They's system containers, so each container runs its own systemd, and in this case, pcscd and yubikey-agent. As long as I make one stop using the USB device, another one can start using it; automatically releasing the USB device on idle would make that mostly happen without user action.

tv42:
Others working on pretty much the same problem (more in the context of letting go of the USB device than between clients of one pcscd): https://github.com/GoogleChromeLabs/chromeos_smart_card_connector/issues/122

ericchiang:
Releasing the smart card context can already be done by calling YubiKey.Close(), right? It's probably better for applications to define what "idle" is instead of this package.

https://pkg.go.dev/github.com/go-piv/piv-go/piv?tab=doc#YubiKey.Close

On a less related note, I'd be happy to add an option to this library to use a SCARD_SHARE_SHARED connection instead of SCARD_SHARE_EXCLUSIVE at the cost of cached PINs. But I think that'll only help multiple applications connecting to the same pcscd.

markomitranic:
Any update on this front guys? :)

ericchiang:
If anyone wants to add a client option to opt into SCARD_SHARE_SHARED I'm happy to review.


https://github.com/go-piv/piv-go/blob/v1.8.0/piv/piv.go#L135

delucca:
Any updates on this?

cedws:
Looks like the PR for piv-go is stuck.

[...]

Issue #45, opened 2020-04-26
Add timeouts or context to methods like Open


ericchiang:
As reported in https://github.com/go-piv/piv-go/issues/42#issuecomment-619481784

PR #51, opened 2020-04-27
piv: mark PC/SC connection as exclusive

ericchiang:
Since this package holds onto a transaction, ensure other calls to Open exit immediately instead of hanging.

Updates #45
Updates #47


ericchiang:
Going to merge this for now since the YubiKey does require an exclusive transaction. If we figure out how to address #47, we can move it back to shared.

PR #57, opened 2020-05-01
add initial windows support

### LINKED ISSUE #56: Support windows via winscard
Inspired by the last stream of @FiloSottile I've digged into piv-go and windows and created a port for windows. Are the maintainers willing to merge windows patches into this repo? If it's fine I would put together a PR and would be willing to work with everyone involved.

Thanks in advance and thanks for this nice library.

tobiaskohlbau:
Adds support for windows via winscard. This CL leverages the
winscard api via calling into the dll. For more information
see https://docs.microsoft.com/en-us/windows/win32/api/winscard/.

Fixes #56

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,11 @@
 .PHONY: test
-test:
+test: lint
 	go test -v ./...
+
+.PHONY: lint
+lint:
 	golint -set_exit_status ./...
+
+.PHONY: build
+build: lint
+	go build ./...
--- a/README.md
+++ b/README.md
@@ -248,6 +248,17 @@ sudo yum config-manager --set-enabled PowerTools
 sudo yum install pcsc-lite-devel
 ```
 
+On Windows:
+
+No prerequisites are needed. The default driver by Microsoft supports all functionalities
+which get tested by unittests. However if you run into problems try the official
+[YubiKey Smart Card Minidriver](https://www.yubico.com/products/services-software/download/smart-card-drivers-tools/). Yubico states on their website the driver adds [_additional
+smart functionality_](https://www.yubico.com/authentication-standards/smart-card/).
+
+Please notice the following:
+
+>Windows support is best effort due to lack of test hardware. This means the maintainers will take patches for Windows, but if you encounter a bug or the build is broken, you may be asked to fix it.
+
 ## Testing
 
 Tests automatically find connected available YubiKeys, but won't modify the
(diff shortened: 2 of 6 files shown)


tobiaskohlbau:
Builders look happy now, sorry for the spam in this commit. The CI-Force is not strong in me 😃.

Makefile
```diff
@@ -1,4 +1,11 @@
-.PHONY: test
+.PHONY: test lint
```
ericchiang, line 1: nit: can you keep this as

```
.PHONY: test
test:
```

I prefer to keep the .PHONY blocks above the rule they apply to.

README.md
```diff
@@ -248,6 +248,15 @@ sudo yum config-manager --set-enabled PowerTools
 sudo yum install pcsc-lite-devel
 ```
 
+On Windows:
+
+No prerequisites are needed. Although it's a good idea to replace the default driver
+with the [YubiKey Smart Card Minidriver](https://www.yubico.com/products/services-software/download/smart-card-drivers-tools/).
```
ericchiang, line 254: Could you elaborate on why is it a good idea? Will users experience errors otherwise?

Makefile
tobiaskohlbau: Wasn't planned this way should have been:

```
.PHONY: test
test: lint
```

similar to the new build task. I've updated the file.

README.md
tobiaskohlbau: I'm was not sure as Yubico states on their website their driver adds additional features like pin change and so on. I've uninstalled the driver and ran the tests with the default driver by Microsoft and everything seems to work. I rephrased the paragraph to state that the default driver looks fine for the unittests but maybe the additional driver is needed for advanced usage.

tobiaskohlbau:
I've addressed your comments. I hope it's fine that I've force pushed. Github is still not working nicely with this workflow, sorry.

ericchiang:
> I've addressed your comments. I hope it's fine that I've force pushed. Github is still not working nicely with this workflow, sorry.

It looks like the windows build worked fine. Am I seeing something different?

https://github.com/go-piv/piv-go/pull/57/checks?check_run_id=638496144

Will merge this including the workflow if you think it's running correctly

tobiaskohlbau:
Oh I get it my comment is confusing. I referred to the workflow of force pushing which deletes the association between the new code and the old comments.

The GitHub actions workflow run is totally fine.

tobiaskohlbau:
I've added a build constraint for `_unix.go` to only build on darwin and linux. Dont know why but I thought unix is supported itself but it's only supported by explicit build constraints as comments.

PR #67, opened 2020-05-17
Initial CHUID reading

jc-m:
Do not merge - work in progress

### DIFF
--- a/piv/piv.go
+++ b/piv/piv.go
@@ -846,3 +846,38 @@ func ykSetProtectedMetadata(tx *scTx, key [24]byte, m *Metadata) error {
 	}
 	return nil
 }
+
+// Card Holder Unique Identifier
+type CardId []byte
+
+func (yk *YubiKey) CardId() (CardId, error) {
+	return ykGetCardId(yk.tx)
+
+}
+func ykGetCardId(tx *scTx) (CardId, error) {
+	// https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-73-4.pdf#page=17
+	// OID for CardId is 5FC102
+
+	cmd := apdu{
+		instruction: insGetData,
+		param1:      0x3f,
+		param2:      0xff,
+		data: []byte{
+			0x5c, // Tag list
+			0x03,
+			0x5f,
+			0xc1,
+			0x02,
+		},
+	}
+	resp, err := tx.Transmit(cmd)
+	if err != nil {
+		return nil, fmt.Errorf("command failed: %w", err)
+	}
+	// https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-73-4.pdf#page=85
+	obj, _, err := unmarshalASN1(resp, 1, 0x13) // tag 0x53
+	if err != nil {
+		return nil, fmt.Errorf("unmarshaling response: %v", err)
+	}
+	return obj, nil
+}


piv/piv.go
```diff
@@ -846,3 +846,121 @@ func ykSetProtectedMetadata(tx *scTx, key [24]byte, m *Metadata) error {
 	}
 	return nil
 }
+
+// CardID is the Card Holder Unique Identifier with settable GUID
+// Raw contains the whole object
+// GUID contains the Card Universally Unique Identifier.
+type CardID struct {
+	Raw  []byte
```
ericchiang, line 854: Make raw unexported and in a future PR add whatever other field you want to be able to inspect. FASC-N, Exp. Date, etc.

Though if you want to use the signature in the API in a follow up PR, I think I'd prefer something like:

```go
func (c *CardID) Verify(pub crypto.PublicKey) error
func (c *CardID) Sign(priv crypto.Signer) error
```

piv/piv.go
```diff
@@ -846,3 +846,121 @@ func ykSetProtectedMetadata(tx *scTx, key [24]byte, m *Metadata) error {
...
+	if err != nil {
+		return id, fmt.Errorf("command failed: %w", err)
+	}
+	// https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-73-4.pdf#page=85
+	obj, _, err := unmarshalASN1(resp, 1, 0x13) // tag 0x53
+	if err != nil {
+		return id, fmt.Errorf("unmarshaling response: %v", err)
+	}
+	id.Raw = obj
+	if obj[27] == 0x34 {
```
ericchiang, line 916: Before doing any slice indexes you need to verify the length of the returned object.

```
if len(obj) < n {
    // return error
}
uuid := obj[n:]
```

piv/piv.go
```diff
@@ -846,3 +846,121 @@ func ykSetProtectedMetadata(tx *scTx, key [24]byte, m *Metadata) error {
...
+		endPos := uuidOffset + int(obj[28])
```
ericchiang, line 917: UUID length must be 16 bytes? https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-73-4.pdf#page=26

Return an error if it's not?

piv/piv.go
```diff
@@ -846,3 +846,121 @@ func ykSetProtectedMetadata(tx *scTx, key [24]byte, m *Metadata) error {
...
+}
+
+// SetCardID initialize the CHUID card object using a predefined template defined as
+//  * Defined fields:
+//    - 0x30: FASC-N (hard-coded)
+//    - 0x34: Card UUID / GUID (settable)
+//    - 0x35: Exp. Date (hard-coded)
+//    - 0x3e: Signature (hard-coded, empty)
+//    - 0xfe: Error Detection Code (hard-coded)
+func (yk *YubiKey) SetCardID(GUID [16]byte, key [24]byte) (CardID, error) {
```
ericchiang, line 930: Can this be:

```
func(yk *YubiKey) SetCardID(key [24], cardID *CardID) error
```

* All other methods in this package have the management key/PIN/PUK as their first argument
* If we ever wanted other methods to be settable (e.g. signature) then we don't need to add new APIs
* We don't return a CardID, a successful SetCardID should imply that the state on the card is equivalent to the passed CardID

piv/piv.go
```diff
@@ -846,3 +846,121 @@ func ykSetProtectedMetadata(tx *scTx, key [24]byte, m *Metadata) error {
...
+// CardID is the Card Holder Unique Identifier with settable GUID
+// Raw contains the whole object
+// GUID contains the Card Universally Unique Identifier.
+type CardID struct {
+	Raw  []byte
+	GUID [16]byte
+}
+
+// CardID returns the card CHUID with the GUID extracted
+func (yk *YubiKey) CardID() (CardID, error) {
```
ericchiang, line 859: For the doc comment: can you call this method before calling SetCardID?

nit: for consistency with other APIs in this package, return a pointer

```
func (yk *YubiKey) CardID() (*CardID, error)
```

piv/piv.go
```diff
@@ -846,3 +846,121 @@ func ykSetProtectedMetadata(tx *scTx, key [24]byte, m *Metadata) error {
...
+	id.Raw = obj
+	if obj[27] == 0x34 {
+		endPos := uuidOffset + int(obj[28])
+		copy(id.GUID[:], obj[uuidOffset:endPos])
+	}
+	return
+}
+
+// SetCardID initialize the CHUID card object using a predefined template defined as
+//  * Defined fields:
```
ericchiang, line 924: The Defined fields is really helpful for review :smiley: but should be in the doc comment. Doc comment should only document what's exposed in the API

piv/piv.go
```diff
@@ -846,3 +846,121 @@ func ykSetProtectedMetadata(tx *scTx, key [24]byte, m *Metadata) error {
...
+ *
+ * Defined fields:
+ *  - 0x30: FASC-N (hard-coded)
+ *  - 0x34: Card UUID / GUID (settable)
+ *  - 0x35: Exp. Date (hard-coded)
+ *  - 0x3e: Signature (hard-coded, empty)
+ *  - 0xfe: Error Detection Code (hard-coded)
+ */
+
+var chuidTemplate = []byte{
```
ericchiang, line 880: Where does this come from? Would default field values be better?

piv/piv.go
```diff
@@ -846,3 +846,121 @@ func ykSetProtectedMetadata(tx *scTx, key [24]byte, m *Metadata) error {
...
+//    - 0x3e: Signature (hard-coded, empty)
+//    - 0xfe: Error Detection Code (hard-coded)
+func (yk *YubiKey) SetCardID(GUID [16]byte, key [24]byte) (CardID, error) {
+	return ykSetCardID(yk.tx, key, GUID)
+
+}
+
+func ykSetCardID(tx *scTx, key [24]byte, guid [16]byte) (id CardID, err error) {
+
+	id.Raw = make([]byte, len(chuidTemplate))
```
ericchiang, line 937: Just build the fields yourself with default values instead of using a template?

```
data := marshalASN1(/* tag */, defaultFASCN)
data = append(data, marshalASN1(/* tag */, guid[:]))
data = append(data, marshalASN1(/* tag */, defaultExp))
data = append(data, marshalASN1(/* tag */, defaultSig))
data = append(data, marshalASN1(/* tag */, defaultErrorCorrectionCode))
```

That way if you ever want to make something like the signature writable, it's clear where to modify the value.

piv/piv.go
```diff
@@ -846,3 +846,121 @@ func ykSetProtectedMetadata(tx *scTx, key [24]byte, m *Metadata) error {
...
+ *
+ * FASC-N containing S9999F9999F999999F0F1F0000000000300001E encoded in
+ * 4-bit BCD with 1 bit parity. run through the tools/fasc.pl script to get
+ * bytes. This CHUID has an expiry of 2030-01-01.
+ *
+ * Defined fields:
+ *  - 0x30: FASC-N (hard-coded)
+ *  - 0x34: Card UUID / GUID (settable)
+ *  - 0x35: Exp. Date (hard-coded)
+ *  - 0x3e: Signature (hard-coded, empty)
```
ericchiang, line 876: Are these actually not settable or is that just the API that yubico's tools expose?

piv/piv.go
jc-m: agreed

piv/piv.go
jc-m: agreed

piv/piv.go
jc-m: can we do that in a future PR as it would be better to pass a full cardID and not have any default ?

piv/piv.go
jc-m: agreed

piv/piv.go
jc-m: agreed

piv/piv.go
```diff
@@ -846,3 +846,121 @@ func ykSetProtectedMetadata(tx *scTx, key [24]byte, m *Metadata) error {
...
+	var id CardID
+	id.raw = obj
+	if len(obj) > uuidOffset {
+		// UUID is a TLV at position 27 with type equal to 0x34
+		if obj[27] == 0x34 {
+			if obj[28] != 0x10 {
+				return nil, fmt.Errorf("unexpected uuid length value: %d", obj[28])
+			}
+			endPos := uuidOffset + int(obj[28])
+			copy(id.GUID[:], obj[uuidOffset:endPos])
```
ericchiang, line 924: endPos can still be longer than the length of obj if the card returns bad data, resulting in a crash.

Can you do the same logic for decoding the metadata instead? Iterate through the raw ASN.1 objects until you find the UUID value?

https://github.com/go-piv/piv-go/blob/e6548dd11f020eb8a3922086893dee86537b47ce/piv/piv.go#L775

ericchiang:
Still plan to take this, but took some extra time off for the long weekend. Will look again when I'm back :)

w6ipa:
any way to get this reviewed/merged to close it ?

jc-m:
Any way to get this reviewed/merged ? Thanks

PR #69, opened 2020-05-18
Add custom Ed25519 signature support with alg 0x22

nickray:
Thoughts on this? My assumption is that the maintenance burden would be minimal :)


nickray:
Sure.

nickray:
Thanks!

ericchiang:
Thanks for the PR :)

PR #72, opened 2020-06-06
Add Retired Key Management Slots

DaKnOb:
This Pull Request adds constants for the Retired Key Management slots of the YubiKey 5 series. This allows the possibility of using more key slots using this library.

### DIFF
--- a/piv/key.go
+++ b/piv/key.go
@@ -244,6 +244,44 @@ var (
 	slotAttestation = Slot{0xf9, 0x5fff01}
 )
 
+var retiredKeyManagementSlots = map[uint32]Slot{
+	0x82: {0x82, 0x5fc10d},
+	0x83: {0x83, 0x5fc10e},
+	0x84: {0x84, 0x5fc10f},
+	0x85: {0x85, 0x5fc110},
+	0x86: {0x86, 0x5fc111},
+	0x87: {0x87, 0x5fc112},
+	0x88: {0x88, 0x5fc113},
+	0x89: {0x89, 0x5fc114},
+	0x8a: {0x8a, 0x5fc115},
+	0x8b: {0x8b, 0x5fc116},
+	0x8c: {0x8c, 0x5fc117},
+	0x8d: {0x8d, 0x5fc118},
+	0x8e: {0x8e, 0x5fc119},
+	0x8f: {0x8f, 0x5fc11a},
+	0x90: {0x90, 0x5fc11b},
+	0x91: {0x91, 0x5fc11c},
+	0x92: {0x92, 0x5fc11d},
+	0x93: {0x93, 0x5fc11e},
+	0x94: {0x94, 0x5fc11f},
+	0x95: {0x95, 0x5fc120},
+}
+
+// RetiredKeyManagementSlot provides access to "retired" slots. Slots meant for old Key Management
+// keys that have been rotated. YubiKeys 4 and later support values between 0x82 and 0x95 (inclusive).
+//
+//     slot, ok := RetiredKeyManagementSlot(0x82)
+//     if !ok {
+//         // unrecognized slot
+//     }
+//     pub, err := yk.GenerateKey(managementKey, slot, key)
+//
+// https://developers.yubico.com/PIV/Introduction/Certificate_slots.html#_slot_82_95_retired_key_management
+func RetiredKeyManagementSlot(key uint32) (Slot, bool) {
+	slot, ok := retiredKeyManagementSlots[key]
+	return slot, ok
+}
+
 // Algorithm represents a specific algorithm and bit size supported by the PIV
 // specification.
 type Algorithm int
--- a/piv/key_test.go
+++ b/piv/key_test.go
@@ -579,3 +579,48 @@ func TestYubiKeyPrivateKeyPINError(t *testing.T) {
 		t.Errorf("expected sign to fail with pin prompt that returned error")
 	}
 }
+
+func TestRetiredKeyManagementSlot(t *testing.T) {
+	tests := []struct {
+		name     string
+		key      uint32
+		wantSlot Slot
+		wantOk   bool
+	}{
+		{
+			name:     "Non-existent slot, before range",
+			key:      0x0,
+			wantSlot: Slot{},
+			wantOk:   false,
+		},
+		{
+			name:     "Non-existent slot, after range",
+			key:      0x96,
+			wantSlot: Slot{},
+			wantOk:   false,
+		},
+		{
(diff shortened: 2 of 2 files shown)


ericchiang:
Hey sorry for the delay here. Didn't have much bandwidth the last few weeks.

So a couple questions with the proposed API:
* It's kind of odd that this is a slice while the others are their own objects, do you expect users to do something like `piv.SlotRetiredKeyManagement[7]`?
* By definition these are retried and going to stop working in the future, what are the caveats here?
* You can already define these yourself. Maybe a comment would be better and projects can use them as they'd like to?

DaKnOb:
It's okay, don't worry, we all do this on our personal free time, there's no expectations or a hurry :)

Indeed, I expect users to use them as you said, mainly because of the amount of them. Being able to iterate over them quickly and easily would help people use them more efficiently.
I think the "Retired" part of this does not mean they are no longer used. They are slots for retired keys. So if a GPG encryption key is rotated for example, the previous one is stored in one of these slots, and the new one in the appropriate slot. This is done so previous e-mails can still be decrypted successfully. This at least is according to my understanding of the documentation. In the Yubikey 5 Series, all these slots are available. Having the ability to access 20 more slots for certificates and keys, with a configurable touch policy and PIN policy for each one of them, in the same hardware device, allows for much greater flexibility for applications that want to use more keys, and it also allows normal "Key Retirement" operations using the library.

I have sent this patch to include this in the library as I believe it will be useful to more than just me. However, if you don't believe they should be included, this is totally fine :)

Thanks for the great work!

ericchiang:
Ah okay, I misunderstood the "retirement" part of the slots. I thought it meant "depreciated" not "store old rotated keys here".

The currently the API is sharp because it's not clear what guarantees it would provide. Are we allow to change the order in the future? Or add more? If we ever add any we could break code that does something like:

```go
// We added a new slot and now this code returns a different value.
lastRetiredSlot := piv.SlotRetiredKeyManagement[len(piv.SlotRetiredKeyManagement)-1]
```

Or if we add new values, would looping over them potentially break old keys?

```go
for _, slot := range piv.SlotRetiredKeyManagement {
    // Woops, we added a new slot and now this loop accesses a slot
    // that the card doesn't support!
}
```

Maybe instead, something like this?

```go
// RetiredKeyManagementSlot provides access to "retired" slots. Slots meant for old Key Management
// keys that have been rotated. YubiKeys 4 and later support values between 0x82 and 0x95 (inclusive).
//
//     slot, ok := RetiredKeyManagementSlot(0x82)
//     if !ok {
//         // unrecognized slot
//     }
//     pub, err := yk.GenerateKey(managementKey, slot, key)
//
// https://developers.yubico.com/PIV/Introduction/Certificate_slots.html#_slot_82_95_retired_key_management
func RetiredKeyManagementSlot(key uint32) (Slot, bool) {
    // ...
}
```

So users have to declare specific slots: 

```go
slot, ok := RetiredKeyManagementSlot(0x82)
```

And can still loop over all the old slots, but in a way we don't break:

```go
var i uint32
for i = 0x82; i <= 0x95; i++ {
    slot, ok := RetiredKeyManagementSlot(i)
    // ...
}
```

DaKnOb:
Indeed this seems like a better idea on how this could be implemented. I was also considering a `map`, however I thought since they were defined in the PIV standard, it was unlikely that they would change, or new ones would be added. I'll make the relevant modifications and let you know.

Thanks for the brainstorming

DaKnOb:
Hello, I've made the necessary modifications and added tests for this, since it's now a function.

piv/key.go
```diff
@@ -244,6 +244,48 @@ var (
...
+//
+//     slot, ok := RetiredKeyManagementSlot(0x82)
+//     if !ok {
+//         // unrecognized slot
+//     }
+//     pub, err := yk.GenerateKey(managementKey, slot, key)
+//
+// https://developers.yubico.com/PIV/Introduction/Certificate_slots.html#_slot_82_95_retired_key_management
+func RetiredKeyManagementSlot(key uint32) (Slot, bool) {
+	var SlotRetiredKeyManagement = []Slot{
```
ericchiang, line 258: Define this globally, not within the function

piv/key.go
```diff
@@ -244,6 +244,48 @@ var (
...
+		{0x90, 0x5fc11b},
+		{0x91, 0x5fc11c},
+		{0x92, 0x5fc11d},
+		{0x93, 0x5fc11e},
+		{0x94, 0x5fc11f},
+		{0x95, 0x5fc120},
+	}
+
+	if key >= 0x82 && key <= 0x95 {
+		return SlotRetiredKeyManagement[key-0x82], true
```
ericchiang, line 282: Just use a map?

```go
val, ok := slotRetiredKeyManagement[key]
if ok {
    return Slot{key, val}, true
}
return Slot{}, false
```

piv/key_test.go
```diff
@@ -579,3 +580,51 @@ func TestYubiKeyPrivateKeyPINError(t *testing.T) {
...
+			name:  "Non-existent slot, after range",
+			args:  args{key: 0x96},
+			want:  Slot{},
+			want1: false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, got1 := RetiredKeyManagementSlot(tt.args.key)
+			if !reflect.DeepEqual(got, tt.want) {
```
ericchiang, line 622: nit: types who's fields are comparable are themselves comparable, you don't need reflection and can just do:

```go
if got != tt.want || gotOK != tt.wantOK {
    t.Errorf(...)
}
```

piv/key_test.go
```diff
@@ -579,3 +580,51 @@ func TestYubiKeyPrivateKeyPINError(t *testing.T) {
...
 	}
 }
+
+func TestRetiredKeyManagementSlot(t *testing.T) {
+	type args struct {
+		key uint32
+	}
+	tests := []struct {
+		name  string
+		args  args
```
ericchiang, line 590: nit: just use "key uint32" instead of a struct here

piv/key_test.go
```diff
@@ -579,3 +580,51 @@ func TestYubiKeyPrivateKeyPINError(t *testing.T) {
...
+		want  Slot
+		want1 bool
```
ericchiang, line 592: nit: "wantOK" instead of "want1"

DaKnOb:
The proposed changes have been implemented and pushed.

ericchiang (approved):
lgtm, please squash and I'll merge :)

DaKnOb:
Done!

PR #74, opened 2020-06-23
fix byte mapping on TouchPolicyAlways and TouchPolicyCached

### LINKED ISSUE #73: incorrect TouchPolicy mapping in the attestation code
The PIV spec says this about the TouchPolicy stored as byte 2 in the KeyPolicy OID:

> 1.3.6.1.4.1.41482.3.8: Two bytes, the first encoding pin policy and the second touch policy
> 
> Pin policy: 01 - never, 02 - once per session, 03 - always
> 
> Touch policy: 01 - never, 02 - always, 03 - cached for 15s

in `key.go` the const's are out of order:

```go
// Touch policies supported by this package.
const (
	TouchPolicyNever TouchPolicy = iota + 1
	TouchPolicyCached
	TouchPolicyAlways
)
```
👆 Cached (2) and Always (3) are reversed. Should be Always=2 and Cached=3

The other spots where touchpolicy parsing/mapping is implemented have the correct values:
```go
var touchPolicyMap = map[TouchPolicy]byte{
	TouchPolicyNever:  0x01,
	TouchPolicyAlways: 0x02,
	TouchPolicyCached: 0x03,
}
```

```go
		case 0x01:
			a.TouchPolicy = TouchPolicyNever
		case 0x02:
			a.TouchPolicy = TouchPolicyAlways
		case 0x03:
			a.TouchPolicy = TouchPolicyCached
```

joemiller:
closes #73

### DIFF
--- a/piv/key.go
+++ b/piv/key.go
@@ -284,8 +284,8 @@ type TouchPolicy int
 // Touch policies supported by this package.
 const (
 	TouchPolicyNever TouchPolicy = iota + 1
-	TouchPolicyCached
 	TouchPolicyAlways
+	TouchPolicyCached
 )
 
 const (


ericchiang:
Cheers!

PR #75, opened 2020-07-11
feat: use pkg-config to discover pcsclite

rawkode:
This was already mentioned in the comments, but not implemented. This allows for `piv-go` to be used successfully as a dependency on Nix builds; as well as being generally more predictable.


piv/pcsc_unix.go
```diff
@@ -18,15 +18,7 @@ package piv
 
 // https://ludovicrousseau.blogspot.com/2010/04/pcsc-sample-in-c.html
 
-// TODO: Figure out if linux flags should use pkg-config instead
-
-// #cgo darwin LDFLAGS: -framework PCSC
```
FiloSottile, line 23: PCSC.framework is part of the OS, we should not be replacing the darwin line.

piv/pcsc_unix.go
```diff
@@ -18,15 +18,7 @@ package piv
...
-// #cgo linux CFLAGS: -I/usr/include/PCSC
-// #cgo linux LDFLAGS: -lpcsclite
-// #cgo freebsd CFLAGS: -I/usr/local/include/
-// #cgo freebsd CFLAGS: -I/usr/local/include/PCSC
-// #cgo freebsd LDFLAGS: -L/usr/local/lib/
-// #cgo freebsd LDFLAGS: -lpcsclite
```
FiloSottile, line 29: Did you test this on FreeBSD?

ericchiang:
To echo what @FiloSottile, please only make changes to platforms you're able to test this on. This doesn't build on MacOS :)

```
% go test -v .
# pkg-config --cflags  -- libpcsclite
pkg-config: exec: "pkg-config": executable file not found in $PATH
FAIL    github.com/go-piv/piv-go/piv [build failed]
FAIL
```

Is it possible to specify pkg-config and fallback to CFLAGS/LDFLAGS if that's not available?

flokli:
Yeah, it seems `pkg-config` and the appropriate `.pc` files aren't available and properly installed on all platforms.

According to https://github.com/golang/go/blob/master/src/go/internal/srcimporter/srcimporter.go#L219-L232, these directives should be additive, and all their outputs are passed to the compiler - So it shouldn't hurt to just add the `pkg-config` line as well with the others.

`#cgo linux {C,LD}FLAGS` could be dropped, if we assume at least all linux distros properly have `pkg-config` set up.

flokli:
@rawkode can you update the PR?

rawkode:
As per feedback, I've updated the PR so that Linux uses `pkg-config` and MacOS and FreeBSD are untouched.

philandstuff:
I’m not sure this works on macOS yet. See https://github.com/NixOS/nixpkgs/pull/93769#issuecomment-663740184 Is there a way to ensure pkg-config is only tried on Linux?

piv/pcsc_unix.go
```diff
@@ -18,14 +18,11 @@ package piv
...
 
-// TODO: Figure out if linux flags should use pkg-config instead
-
 // #cgo darwin LDFLAGS: -framework PCSC
-// #cgo linux CFLAGS: -I/usr/include/PCSC
-// #cgo linux LDFLAGS: -lpcsclite
-// #cgo freebsd CFLAGS: -I/usr/local/include/
-// #cgo freebsd CFLAGS: -I/usr/local/include/PCSC
-// #cgo freebsd LDFLAGS: -L/usr/local/lib/
+// #cgo pkg-config: libpcsclite
```
philandstuff, line 22:
```suggestion
// #cgo linux pkg-config: libpcsclite
```
I have tested this change on linux and macOS and it still compiles correctly on linux while preventing the errors I was getting on macOS.

piv/pcsc_unix.go
```diff
@@ -18,14 +18,11 @@ package piv
...
-// TODO: Figure out if linux flags should use pkg-config instead
-
 // #cgo darwin LDFLAGS: -framework PCSC
-// #cgo linux CFLAGS: -I/usr/include/PCSC
-// #cgo linux LDFLAGS: -lpcsclite
-// #cgo freebsd CFLAGS: -I/usr/local/include/
-// #cgo freebsd CFLAGS: -I/usr/local/include/PCSC
-// #cgo freebsd LDFLAGS: -L/usr/local/lib/
+// #cgo linux pkg-config: libpcsclite
+// #cgo freebsd CFLAGS: -I/usr/local/include/
```
ericchiang, line 23: There's trailing whitespace on these lines, can you remove that so there's no diff for the freebsd cgo lines?

rawkode:
@ericchiang Sorted

Thanks for everyone's help, @flokli @philandstuff @FiloSottile

PR #78, opened 2020-08-23
Remove confusing word from README

psanford:
Previously the README said this was a 'pure Go' implementation
which usually implies no cgo. Since this package wraps c libraries
(pcsc, etc.) this was a bit confusing. Remove the word 'pure'
to avoid that confusion.


ericchiang:
Agreed. Thanks for the PR :)

PR #80, opened 2020-09-04
Support ECDH for ECDSA keys

### LINKED ISSUE #79: ECDH for using P-384 keys for more than signing
go-ykpiv's CGo wrapper of libykpiv has this:
 
https://godoc.org/github.com/go-piv/go-ykpiv#Slot.Decrypt

> f the slot holds an EC key, then we will perform ECDH and return the shared secret. In this case, msg must be the peer's public key in octet form, as specified in section 4.3.6 of ANSI X9.62.

I've tested that to work. I would, however, like to use pure-Go libraries.

I couldn't figure out how to get ECDH working with piv-go. For the CGo version, sign and decrypt look very similar:

https://github.com/go-piv/go-ykpiv/blob/32315ce599d8e21e91880e6ecdce55a98420ddc0/decrypt.go#L68  
https://github.com/go-piv/go-ykpiv/blob/32315ce599d8e21e91880e6ecdce55a98420ddc0/sign.go#L166

and so do libykpiv's two functions, they both call the same function:

https://github.com/Yubico/yubico-piv-tool/blob/0b27308560ab4f3df4bc29a179e1c91649ce0629/lib/ykpiv.c#L1144  
https://github.com/Yubico/yubico-piv-tool/blob/0b27308560ab4f3df4bc29a179e1c91649ce0629/lib/ykpiv.c#L1126

go-ykpiv has a helpful test case here:

https://github.com/go-piv/go-ykpiv/blob/32315ce599d8e21e91880e6ecdce55a98420ddc0/ykpiv_test.go#L522

But I can't see where exactly signing and decrypting are different to make ECDH work with go-piv. Any thoughts?

tv42:
Feedback on APIs would be most welcome.

Fixes #79

### DIFF
--- a/piv/key.go
+++ b/piv/key.go
@@ -31,6 +31,10 @@ import (
 	"math/big"
 )
 
+// errMismatchingAlgorithms is returned when a cryptographic operation
+// is given keys using different algorithms.
+var errMismatchingAlgorithms = errors.New("mismatching key algorithms")
+
 // Slot is a private key and certificate combination managed by the security key.
 type Slot struct {
 	// Key is a reference for a key type.
@@ -657,7 +661,7 @@ func (yk *YubiKey) PrivateKey(slot Slot, public crypto.PublicKey, auth KeyAuth) 
 
 	switch pub := public.(type) {
 	case *ecdsa.PublicKey:
-		return &keyECDSA{yk, slot, pub, auth, pp}, nil
+		return &ECDSAPrivateKey{yk, slot, pub, auth, pp}, nil
 	case ed25519.PublicKey:
 		return &keyEd25519{yk, slot, pub, auth, pp}, nil
 	case *rsa.PublicKey:
@@ -667,7 +671,13 @@ func (yk *YubiKey) PrivateKey(slot Slot, public crypto.PublicKey, auth KeyAuth) 
 	}
 }
 
-type keyECDSA struct {
+// ECDSAPrivateKey is a crypto.PrivateKey implementation for ECDSA
+// keys. It implements crypto.Signer and the method SharedKey performs
+// Diffie-Hellman key agreements.
+//
+// Keys returned by YubiKey.PrivateKey() may be type asserted to
+// *ECDSAPrivateKey, if the slot contains an ECDSA key.
+type ECDSAPrivateKey struct {
 	yk   *YubiKey
 	slot Slot
 	pub  *ecdsa.PublicKey
@@ -675,13 +685,69 @@ type keyECDSA struct {
 	pp   PINPolicy
 }
 
-func (k *keyECDSA) Public() crypto.PublicKey {
+// Public returns the public key associated with this private key.
+func (k *ECDSAPrivateKey) Public() crypto.PublicKey {
 	return k.pub
 }
 
-func (k *keyECDSA) Sign(rand io.Reader, digest []byte, opts crypto.SignerOpts) ([]byte, error) {
+var _ crypto.Signer = (*ECDSAPrivateKey)(nil)
+
+// Sign implements crypto.Signer.
+func (k *ECDSAPrivateKey) Sign(rand io.Reader, digest []byte, opts crypto.SignerOpts) ([]byte, error) {
 	return k.auth.do(k.yk, k.pp, func(tx *scTx) ([]byte, error) {
 		return ykSignECDSA(tx, k.slot, k.pub, digest)
+	})
+}
+
+// SharedKey performs a Diffie-Hellman key agreement with the peer
+// to produce a shared secret key.
+//
+// Peer's public key must use the same algorithm as the key in
+// this slot, or an error will be returned.
+//
+// Length of the result depends on the types and sizes of the keys
+// used for the operation. Callers should use a cryptographic key
+// derivation function to extract the amount of bytes they need.
+func (k *ECDSAPrivateKey) SharedKey(peer *ecdsa.PublicKey) ([]byte, error) {
+	if peer.Curve.Params().BitSize != k.pub.Curve.Params().BitSize {
+		return nil, errMismatchingAlgorithms
+	}
+	msg := elliptic.Marshal(peer.Curve, peer.X, peer.Y)
+	return k.auth.do(k.yk, k.pp, func(tx *scTx) ([]byte, error) {
+		var alg byte
+		size := k.pub.Params().BitSize
+		switch size {
+		case 256:
+			alg = algECCP256
+		case 384:
+			alg = algECCP384
+		default:
+			return nil, fmt.Errorf("unsupported curve: %d", size)
+		}
+
+		// https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-73-4.pdf#page=118
+		// https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-73-4.pdf#page=93
+		cmd := apdu{
+			instruction: insAuthenticate,
+			param1:      alg,
+			param2:      byte(k.slot.Key),
+			data: marshalASN1(0x7c,
+				append([]byte{0x82, 0x00},
+					marshalASN1(0x85, msg)...)),
+		}
+		resp, err := tx.Transmit(cmd)
+		if err != nil {
+			return nil, fmt.Errorf("command failed: %w", err)
+		}
+		sig, _, err := unmarshalASN1(resp, 1, 0x1c) // 0x7c
+		if err != nil {
+			return nil, fmt.Errorf("unmarshal response: %v", err)
+		}
+		rs, _, err := unmarshalASN1(sig, 2, 0x02) // 0x82
+		if err != nil {
+			return nil, fmt.Errorf("unmarshal response signature: %v", err)
+		}
+		return rs, nil
 	})
 }
 
--- a/piv/key_test.go
+++ b/piv/key_test.go
@@ -78,6 +78,65 @@ func TestYubiKeySignECDSA(t *testing.T) {
 	}
 }
 
+func TestYubiKeyECDSASharedKey(t *testing.T) {
+	yk, close := newTestYubiKey(t)
+	defer close()
+
+	slot := SlotAuthentication
+
+	key := Key{
+		Algorithm:   AlgorithmEC256,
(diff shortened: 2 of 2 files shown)


piv/key.go
```diff
@@ -31,6 +31,10 @@ import (
 	"math/big"
 )
 
+// ErrMismatchingAlgorithms is returned when a cryptographic operation
+// is given keys using different algorithms.
+var ErrMismatchingAlgorithms = errors.New("mismatching key algorithms")
```
ericchiang, line 36: I wouldn't export this of even use a variable here. Just have the test determine if this returns an error or not.

piv/key_test.go
```diff
@@ -78,6 +78,80 @@ func TestYubiKeySignECDSA(t *testing.T) {
 	}
 }
 
+func TestYubiKeyECDSAKeyAgreement(t *testing.T) {
+	yk, close := newTestYubiKey(t)
+	defer close()
+
+	if err := yk.Reset(); err != nil {
```
ericchiang, line 85: nit: I don't think this test needs to reset the key (that takes a while). Calling GenerateKey() should overwrite any existing content.

ericchiang (commented):
Thanks for sending a PR :)

Keys implement an optional interface because it's what the standard library wants. The only prior art I could find for Diffie-Hellman was:

https://github.com/golang/go/blob/01a9cf8487df2b108f0dfd7060ff5ffbda972c3a/src/crypto/tls/key_schedule.go#L104-L110

I think this package shouldn't establishing a key agreement interface. Maybe just export keyECDSA as concrete type that you can type switch on?

```go
type ECPrivetKey struct {}

func (k *ECPrivateKey) Public() crypto.Public {}
func (k *ECPrivateKey) Sign(rand io.Reader, digest []byte, opts crypto.SignerOpts) ([]byte, error)
func (k *ECPrivateKey) KeyAgreement(peer *ecdsa.PublicKey) ([]byte, error)
```

Then using the API looks like:

```go
priv, err := yk.PrivateKey(slot, pub, auth)
if err != nil {
    // ...
}
ecPriv, ok := priv.(*piv.ECPrivateKey)
if !ok {
    // ...
}
sharedKey, err := ecPriv.KeyAgreement(peer)
```

If there's ever an actual key agreement interface to satisfy in the future, then ECPrivateKey can implement that later.

piv/key.go
```diff
@@ -727,7 +761,9 @@ func (k *keyRSA) Decrypt(rand io.Reader, msg []byte, opts crypto.DecrypterOpts)
 	})
 }
 
-func ykSignECDSA(tx *scTx, slot Slot, pub *ecdsa.PublicKey, digest []byte) ([]byte, error) {
+func ykSignECDSAOrECDH(tx *scTx, slot Slot, pub *ecdsa.PublicKey, op byte, msg []byte) ([]byte, error) {
```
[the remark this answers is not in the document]
tv42: Sure, I can copy-paste. They were so very similar that I figured I'd try to share. I was following ykpiv to see what the difference is, and it has a single helper function called by both; I had no clue on whether there would ever be more than two callers of that helper function.

piv/key.go
tv42: Okay.

piv/key_test.go
tv42: Sweet. I was stumbling in the dark there.

tv42:
Ooh I hadn't found the stdlib one. Of course it makes sense TLS would have one. I'll happily steal the `SharedKey` name for that.

The reason I exported an interface, not a concrete type, was that I was thinking there might be a non-ECDSA keytype that can do Diffie-Hellman on the yubikey. But yeah the interface is easy to add later, if and when. I can switch to a concrete type.

tv42:
Naming question: You used `ECPrivateKey`. Since this is specifically a P-nnn key that is ECDSA, and will refuse to do DH with non-ECDSA keys, I think it should be `ECDSAPrivateKey` then. We're not speaking on behalf of all elliptic curve crypto. Callers can always use handle multiple concrete types with interfaces.

ericchiang:
ECDSAPrivateKey works for me

tv42:
Pushed fixups for all feedback above. If you're happy with the end result, I'll rebase them together and take the draft flag off this PR.

piv/key.go
```diff
@@ -667,24 +671,82 @@ func (yk *YubiKey) PrivateKey(slot Slot, public crypto.PublicKey, auth KeyAuth)
...
+// SharedKey performs a Diffie-Hellman key agreement with the peer
+// to produce a shared secret key.
+//
+// Peer's public key must use the same algorithm as the key in
+// this slot, or an error will be returned.
+//
+// Length of the result depends on the types and sizes of the keys
+// used for the operation. Callers should use a cryptographic key
+// derivation function to extract the amount of bytes they need.
+func (k *ECDSAPrivateKey) SharedKey(rand io.Reader, peer crypto.PublicKey, opts crypto.SignerOpts) ([]byte, error) {
```
ericchiang, line 704: Unless there's a good reason to keep rand, crypto.PublicKey, and opt, I'd rather drop the unused arguments and have more type safety here.

```
func (k *ECDSAPrivateKey) SharedKey(peer *ecdsa.PublicKey) ([]byte, error) {}
```

piv/key.go
```diff
@@ -667,24 +671,82 @@ func (yk *YubiKey) PrivateKey(slot Slot, public crypto.PublicKey, auth KeyAuth)
...
+		case 256:
+			alg = algECCP256
+		case 384:
+			alg = algECCP384
+		default:
+			return nil, fmt.Errorf("unsupported curve: %d", size)
+		}
+
+		// https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-73-4.pdf#page=118
+		cmd := apdu{
```
ericchiang, line 726: nit: can you also add a comment linking to

https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-73-4.pdf#page=93

That's what specifies 0x85.

piv/key.go
```diff
@@ -667,24 +671,82 @@ func (yk *YubiKey) PrivateKey(slot Slot, public crypto.PublicKey, auth KeyAuth)
 	}
 }
 
-type keyECDSA struct {
+// ECDSAPrivateKey is used to access a ECDSA signing key.
```
ericchiang, line 674: nit: include user instructions to get a *ECDSAPrivateKey. Anyone reading the docs will probably appreciate that :)

```
// ECDSAPrivateKey is a crypto.PrivateKey implementation used to directly access ECDSA
// operations. Keys returned by PrivateKey() can be type casted to use these methods.
//
//     priv, err := yk.PrivateKey(slot, pub, auth)
//     if err != nil {
//         // ...
//     }
//     ecdsaPriv, ok := priv.(piv. ECDSAPrivateKey)
```

ericchiang (approved):
Couple comments, but other than that lgtm. Thanks for this!

ericchiang:
Oh an please squash your commits :) I can merge after

piv/key.go
tv42: Yeah. I have no idea what the expectation is. These seem to be offered "in case the algorithm needs them" all around the PrivateKey/PublicKey ecosystem, which makes sense for interfaces that multiple algorithms would offer.

Worst case scenario here is that piv-go ends up with a `SharedKey` method that looks different from a hypothetical future stdlib interface. It might be awkward, but it's always possible to write an adapter that drops them, in front of this type, or do a major version bump to change the implementation.

I'll rip out the unused arguments.

piv/key.go
tv42: I wrote proper examples for SharedKey, that should also demonstrate this, and the documentation for the type referring there.

tv42:
Squashed, all ready to go. (Branch `wip-ecdh` has the commit-by-commit action, if you want to see the latest changes.)

tv42:
Oops, sorry, I want to clean up one more nit in the example.

tv42:
Nope, nevermind, what I needed to do was refresh my godoc window to see that I had already fixed the nit. All good!

piv/example_test.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rand"
+	"log"
+
+	"github.com/go-piv/piv-go/piv"
+)
+
+func ExampleECDSAPrivateKey_SharedKey_epheremalEncrypt() {
```
ericchiang, line 26: Couple issues I see with these examples:
* There's no encrypt or decryption here as the name would suggest.
* I don't want to give recommendations on the best ways to use a shared secret for encryption/decryption. There's a lot of ways to get a handshake like this wrong and I'm not a cryptographer.
* The "ExampleECDSAPrivateKey_SharedKey_epheremalEncrypt" doesn't seem like it should belong here, there's nothing piv specific about it.
* Instead of hardwarePublicKey or ephemeralPublicKey, these should probably be publicKey and peerPublicKey.

Can I strongly recommend dropping the examples and doing this as a follow up? It's hard to review PRs that keep adding new code during the review :)

piv/key.go
```diff
@@ -667,24 +671,91 @@ func (yk *YubiKey) PrivateKey(slot Slot, public crypto.PublicKey, auth KeyAuth)
...
+// SharedKey performs a Diffie-Hellman key agreement with the peer
+// to produce a shared secret key.
+//
+// Peer's public key must use the same algorithm as the key in
+// this slot, or an error will be returned.
+//
+// Length of the result depends on the types and sizes of the keys
+// used for the operation. Callers should use a cryptographic key
+// derivation function to extract the amount of bytes they need.
+func (k *ECDSAPrivateKey) SharedKey(peer crypto.PublicKey) ([]byte, error) {
```
ericchiang, line 712: Didn't see this addressed in the comments, is there a reason you kept this as crypto.PublicKey instead of *ecdsa.PublicKey? I'd prefer the latter since it enforces type safety.

tv42: The downside of strong type there is that will never match any future interfaces defined. But that can be sorted if and when such an interface arises.

piv/example_test.go
tv42: Dropped examples.

ericchiang (approved):

Issue #77, opened 2020-08-07
implement yubikey SetKey similar to SetCertificate?


colemickens:
If I managed to get the APDU commands working for yubikeys to import their private key, would you accept a PR for that functionality?

ericchiang:
Been a while Cole 😀 hope you're doing well!

The reason I've omitted this feature is because I think it's a bit of an anti-pattern. You should be able to attest a key is in hardware, or there's not a huge benefit in having hardware backed keys.

So, I don't understand the use case of importing a key. If you've gotten some ideas happy to hear them, but otherwise I'd like to avoid implementing loading private keys.

ericchiang:
Closed by #83

PR #83, opened 2020-12-02
Implement import key functionality

hayleyjames:
Issue #77 discusses whether this library should support this feature. My use case for implementing this is to be able to create backup Yubikeys with the same private key.

Tested on: Ubuntu 20.04.1 LTS

### DIFF
--- a/piv/key.go
+++ b/piv/key.go
@@ -35,6 +35,18 @@ import (
 // is given keys using different algorithms.
 var errMismatchingAlgorithms = errors.New("mismatching key algorithms")
 
+// errUnsupportedKeySize is returned when a key has an unsupported size
+var errUnsupportedKeySize = errors.New("unsupported key size")
+
+// unsupportedCurveError is used when a key has an unsupported curve
+type unsupportedCurveError struct {
+	curve int
+}
+
+func (e unsupportedCurveError) Error() string {
+	return fmt.Sprintf("unsupported curve: %d", e.curve)
+}
+
 // Slot is a private key and certificate combination managed by the security key.
 type Slot struct {
 	// Key is a reference for a key type.
@@ -445,19 +457,25 @@ func (yk *YubiKey) Certificate(slot Slot) (*x509.Certificate, error) {
 	return cert, nil
 }
 
-// marshalASN1 encodes a tag, length and data.
-//
-// TODO: clean this up and maybe switch to cryptobyte?
-func marshalASN1(tag byte, data []byte) []byte {
+// marshalASN1Length encodes the length.
+func marshalASN1Length(n uint64) []byte {
 	var l []byte
-	n := uint64(len(data))
 	if n < 0x80 {
 		l = []byte{byte(n)}
-	} else if len(data) < 0x100 {
+	} else if n < 0x100 {
 		l = []byte{0x81, byte(n)}
 	} else {
 		l = []byte{0x82, byte(n >> 8), byte(n)}
 	}
+
+	return l
+}
+
+// marshalASN1 encodes a tag, length and data.
+//
+// TODO: clean this up and maybe switch to cryptobyte?
+func marshalASN1(tag byte, data []byte) []byte {
+	l := marshalASN1Length(uint64(len(data)))
 	d := append([]byte{tag}, l...)
 	return append(d, data...)
 }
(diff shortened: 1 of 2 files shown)


ericchiang (commented):
For consistency this should probably be "SetPrivateKey"

Since there's already a struct that holds the policies, can the arguments be:

```go
func (yk *YubiKey) SetPrivateKey(key [24]byte, slot Slot, priv crypto.PrivateKey, policy Key)
```

Or we can refactor this to:

```go
type KeyPolicy struct {
    PINPolicy PINPolicy
    TouchPolicy TouchPolicy
    // Opt-in required to use imported private keys?
    InsecureAllowImportedPrivateKeys bool
}
func (yk *YubiKey) SetPrivateKey(key [24]byte, slot Slot, priv crypto.PrivateKey, policy KeyPolicy)
```

I can't review the rest without a link to the spec. You can see a few comments that have links to pages in the NIST pdf. Can you include the one that defines the key parameters:

```
https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-73-4.pdf#page=95
```

Finally, I think there should be warnings associated with this method:

```
// SetPrivateKey is an insecure method which imports a private key into the slot.
// Users should almost always use GeneratePrivateKey() instead.
//
// Importing a private key breaks functionality provided by this package, including
// AttestationCertificate() and Attest(). There are no stability guarantees for other
// methods for imported private keys.
//
// Keys generated outside of the YubiKey should not be considered hardware-backed,
// as there's no way to prove the key wasn't copied, exfiltrated, or replaced with malicious
// material before being imported.
func (yk *YubiKey) SetPrivateKey(...)
```

ericchiang:
Also, wanted to say thanks for the PR! Apologies for some if some of the "insecure" comments seem dramatic. For security relevant code, I always find it helpful to optimize for reviewers (so they can see an "Insecure..." variable) and developers who are approaching this for the first time.

hayleyjames:
Thank you for reviewing my PR!

Adding an insecure warning is a good idea. The command is a Yubico extension to PIV. You can find the yubico-piv-tool implementation [here](https://github.com/Yubico/yubico-piv-tool/blob/671a5740ef09d6c5d9d33f6e5575450750b58bde/lib/ykpiv.c#L1812).

I don't think InsecureAllowImportedPrivateKeys flag would work because we can't differentiate between an empty slot and an imported key.

Would naming the function SetPrivateKeyInsecure be appropriate?

piv/key.go
```diff
@@ -709,6 +718,125 @@ func (yk *YubiKey) PrivateKey(slot Slot, public crypto.PublicKey, auth KeyAuth)
...
+		case 2048:
+			policy.Algorithm = AlgorithmRSA2048
+			elemLen = 128
+		default:
+			return errUnsupportedKeySize
+		}
+
+		// P value
+		params = append(params, priv.Primes[0].Bytes())
+
```
ericchiang, line 753: nit: can you remove some of the whitespace here?

```
params = append(params, priv.Primes[0].Bytes()) // P
params = append(params, priv.Primes[1].Bytes()) // Q
params = append(params, priv.Precomputed.Dp.Bytes()) // dP
params = append(params, priv.Precomputed.Dq.Bytes()) // dQ
params = append(params, priv.Precomputed.Qinv.Bytes()) // Qinv
```

piv/key.go
```diff
@@ -709,6 +718,125 @@ func (yk *YubiKey) PrivateKey(slot Slot, public crypto.PublicKey, auth KeyAuth)
...
+		// Q value
+		params = append(params, priv.Primes[1].Bytes())
+
+		// dP value
+		params = append(params, priv.Precomputed.Dp.Bytes())
```
ericchiang, line 758: Precompute values aren't guaranteed to present. Please add a call Precompute before using them:
```
priv.Precompute()
```
Precompute is no-op if the values have already been precomputed https://golang.org/src/crypto/rsa/rsa.go?s=14128:14164#L448

piv/key.go
```diff
@@ -709,6 +718,125 @@ func (yk *YubiKey) PrivateKey(slot Slot, public crypto.PublicKey, auth KeyAuth)
...
+		case 384:
+			policy.Algorithm = AlgorithmEC384
+			elemLen = 48
+		default:
+			return fmt.Errorf("unsupported curve: %d", size)
+		}
+
+		// S value
+		privateKey := make([]byte, elemLen)
+		params = append(params, priv.D.FillBytes(privateKey))
```
ericchiang, line 781: nit: FillBytes is Go 1.15+ but we probably still want to support Go 1.14 until Go 1.16 is out. Can you please replace this with a different method?

https://golang.org/pkg/math/big/#Int.FillBytes
https://golang.org/doc/devel/release.html#policy

piv/key.go
```diff
@@ -709,6 +718,125 @@ func (yk *YubiKey) PrivateKey(slot Slot, public crypto.PublicKey, auth KeyAuth)
...
+//
+// Importing a private key breaks functionality provided by this package, including
+// AttestationCertificate() and Attest(). There are no stability guarantees for other
+// methods for imported private keys.
+//
+// Keys generated outside of the YubiKey should not be considered hardware-backed,
+// as there's no way to prove the key wasn't copied, exfiltrated, or replaced with malicious
+// material before being imported.
+func (yk *YubiKey) SetPrivateKeyInsecure(key [24]byte, slot Slot, private crypto.PrivateKey, policy Key) error {
+	params := make([][]byte, 0)
```
ericchiang, line 732: Can you add a comment linking to https://github.com/Yubico/yubico-piv-tool/blob/671a5740ef09d6c5d9d33f6e5575450750b58bde/lib/ykpiv.c#L1812 here?

piv/key_test.go
```diff
@@ -683,3 +683,166 @@ func TestRetiredKeyManagementSlot(t *testing.T) {
...
+			generated, err := rsa.GenerateKey(rand.Reader, tt.bits)
+			if err != nil {
+				t.Fatalf("generating private key: %v", err)
+			}
+
+			err = yk.SetPrivateKeyInsecure(DefaultManagementKey, tt.slot, generated, Key{
+				PINPolicy:   PINPolicyNever,
+				TouchPolicy: TouchPolicyNever,
+			})
+			if err != nil {
```
ericchiang, line 735: It looks like you're missing if you didn't get an error but wanted one. Suggest changing this to

```go
if (err != nil) != tt.wantErr {
    t.Fatalf("SetPrivateKeyInsecure(): wantErr=%v, got err=%v", tt.wantErr, err)
}
if err != nil {
    return
}
```

piv/key_test.go
```diff
@@ -683,3 +683,166 @@ func TestRetiredKeyManagementSlot(t *testing.T) {
...
+			generated, err := ecdsa.GenerateKey(tt.curve, rand.Reader)
+			if err != nil {
+				t.Fatalf("generating private key: %v", err)
+			}
+
+			err = yk.SetPrivateKeyInsecure(DefaultManagementKey, tt.slot, generated, Key{
+				PINPolicy:   PINPolicyNever,
+				TouchPolicy: TouchPolicyNever,
+			})
+			if err != nil {
```
ericchiang, line 818: nit: same comment as the other condition, might be easier to do:

```
if (err != nil) != tt.wantErr {
    t.Fatalf("SetPrivateKeyInsecure(): wantErr=%v, got err=%v", tt.wantErr, err)
}
if err != nil {
    return
}
```

ericchiang:
"SetPrivateKeyInsecure" works for me

ericchiang:
Tests passed on my Macbook.

lgtm. Can you squash your changes into a single commit? I'll merge after.

https://stackoverflow.com/a/5201642

hayleyjames:
Thanks for reviewing my changes! The tests also pass on Windows 10.

ericchiang:
Thanks for your contribution! This has been included in the v1.7.0 tag https://github.com/go-piv/piv-go/releases/tag/v1.7.0

PR #84, opened 2021-01-07
Improve KeyAuth.PINPolicy documentation regarding imported keys

jstasiak:


piv/key.go
```diff
@@ -618,6 +618,9 @@ type KeyAuth struct {
 	// not provided, this will be inferred from the attestation certificate.
 	//
 	// This field is required on older (<4.3.0) YubiKeys when using PINPrompt.
+	// It is also required if the key has been imported to the card instead of
```
ericchiang, line 621: nit: can this just be:

```
// This field is required on older (<4.3.0) YubiKeys when using PINPrompt,
// as well as keys imported through SetPrivateKeyInsecure().
```

ericchiang:
Thanks! Had a small suggestion on the wording since I don't think we need to go into details.

piv/key.go
jstasiak: Absolutely, I added too many details possibly. I wanted to not mention `SetPrivateKeyInsecure()` by name, since the key could be imported by another library/API/imlementation. How about this?

```
// This field is required on older (<4.3.0) YubiKeys when using PINPrompt,
// as well as for keys imported to the card.
```

I think it's also worth clarifying the situation by mentioning "generated on the card" so that it's clearer to the reader what the distinction is:

```
// This field is required on older (<4.3.0) YubiKeys when using PINPrompt,
// as well as for keys imported to (not generated on) the card.
```

*but* I'm happy with both of those as they're an improvement still, let me know which one do you want.

ericchiang:
```
// This field is required on older (<4.3.0) YubiKeys when using PINPrompt,
// as well as for keys imported to the card.
```
lgtm

jstasiak: Done.

ericchiang:
Thanks! Please squash your commits and I'll merge

PR #89, opened 2021-07-05
Add slot name to Attestation struct

jalseth:
In some use cases, knowing what slot the attested key resides in
can be useful for determining whether to issue a certificate for
the key.

Signed-off-by: James Alseth <james@yubico.com>

### DIFF
--- a/piv/key.go
+++ b/piv/key.go
@@ -29,6 +29,8 @@ import (
 	"fmt"
 	"io"
 	"math/big"
+	"strconv"
+	"strings"
 )
 
 // errMismatchingAlgorithms is returned when a cryptographic operation
@@ -86,8 +88,13 @@ const (
 	FormfactorUSBCLightningKeychain
 )
 
-// Attestation returns additional information about a key attested to be on a
-// card.
+// Prefix in the x509 Subject Common Name for YubiKey attestations
+// https://developers.yubico.com/PIV/Introduction/PIV_attestation.html
+const yubikeySubjectCNPrefix = "YubiKey PIV Attestation "
+
+// Attestation returns additional information about a key attested to be generated
+// on a card. See https://developers.yubico.com/PIV/Introduction/PIV_attestation.html
+// for more information.
 type Attestation struct {
 	// Version of the YubiKey's firmware.
 	Version Version
@@ -102,6 +109,11 @@ type Attestation struct {
 	PINPolicy PINPolicy
 	// TouchPolicy set on the slot.
 	TouchPolicy TouchPolicy
+
+	// Slot is the inferred slot the attested key resides in based on the
+	// common name in the attestation. If the slot cannot be determined,
+	// this field will be an empty struct.
+	Slot Slot
 }
 
 func (a *Attestation) addExt(e pkix.Extension) error {
@@ -210,7 +222,38 @@ func parseAttestation(slotCert *x509.Certificate) (*Attestation, error) {
 			return nil, fmt.Errorf("parsing extension: %v", err)
 		}
 	}
+
+	slot, ok := parseSlot(slotCert.Subject.CommonName)
+	if ok {
+		a.Slot = slot
+	}
+
 	return &a, nil
+}
+
+func parseSlot(commonName string) (Slot, bool) {
+	if !strings.HasPrefix(commonName, yubikeySubjectCNPrefix) {
+		return Slot{}, false
+	}
+
+	slotName := strings.TrimPrefix(commonName, yubikeySubjectCNPrefix)
+	key, err := strconv.ParseUint(slotName, 16, 32)
+	if err != nil {
+		return Slot{}, false
+	}
+
+	switch uint32(key) {
+	case SlotAuthentication.Key:
+		return SlotAuthentication, true
+	case SlotSignature.Key:
+		return SlotSignature, true
+	case SlotCardAuthentication.Key:
+		return SlotCardAuthentication, true
+	case SlotKeyManagement.Key:
+		return SlotKeyManagement, true
+	}
+
+	return RetiredKeyManagementSlot(uint32(key))
 }
 
 // yubicoPIVCAPEM is the PEM encoded attestation certificate used by Yubico.
@@ -296,6 +339,11 @@ var retiredKeyManagementSlots = map[uint32]Slot{
 func RetiredKeyManagementSlot(key uint32) (Slot, bool) {
 	slot, ok := retiredKeyManagementSlots[key]
 	return slot, ok
+}
+
+// String returns the two-character hex representation of the slot
+func (s Slot) String() string {
+	return strconv.FormatUint(uint64(s.Key), 16)
 }
 
 // Algorithm represents a specific algorithm and bit size supported by the PIV
--- a/piv/key_test.go
+++ b/piv/key_test.go
@@ -433,6 +433,12 @@ func TestYubiKeyAttestation(t *testing.T) {
 	if a.Version != yk.Version() {
 		t.Errorf("attestation version got=%#v, wanted=%#v", a.Version, yk.Version())
 	}
+	if a.Slot != SlotAuthentication {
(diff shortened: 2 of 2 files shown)


piv/key.go
```diff
@@ -102,6 +103,9 @@ type Attestation struct {
 	PINPolicy PINPolicy
 	// TouchPolicy set on the slot.
 	TouchPolicy TouchPolicy
+
+	// Slot is the name of the slot the attested key resides in
```
ericchiang, line 107: nit: add a period to the end of the comment.

What are the set of names this could be? "sign" or a hex number? Can you add an example to the comment?

piv/key.go
```diff
@@ -210,6 +214,9 @@ func parseAttestation(slotCert *x509.Certificate) (*Attestation, error) {
 			return nil, fmt.Errorf("parsing extension: %v", err)
 		}
 	}
+	subjectSplit := strings.Split(slotCert.Subject.CommonName, " ")
```
ericchiang, line 217: Why the split? Is this documented anywhere? Can you link to that documentation in a comment? :)

Also, please add a bounds check:

```
if len(subjectSplit) > 0 {
  a.Slot = subjectSplit[len(subjectSplit)]
}
```

ericchiang (commented):
Sorry for the delay. Holidays :)

jalseth:
@ericchiang No need to apologize, a review in under a day is very quick! I believe I've addressed your feedback, and removed the split altogether.

piv/key.go
```diff
@@ -102,6 +103,11 @@ type Attestation struct {
 	PINPolicy PINPolicy
 	// TouchPolicy set on the slot.
 	TouchPolicy TouchPolicy
+
+	// Slot is the two-letter name of the slot (9a, 9c, etc.) the attested key resides in.
+	// See https://developers.yubico.com/PIV/Introduction/Certificate_slots.html for more
+	// information.
+	Slot string
```
ericchiang, line 110: We already have a slot object. Would you want to make this an actual type, rather than just a string?

```
// Inferred Slot based on the common name of the certificate. If the slot can't be determined, this
// field will be equivalent to an empty Slot struct.
Slot Slot
```

We could also give Slot a String() function to make printing it easier

```
func (s *Slot) String() string {
   return strconv.FormatUint(uint64(s.Key), 16)
}
```

Parsing logic then becomes something like:

```
const yubikeySubjectCNPrefix = "YubiKey PIV Attestation "

func parseSlot(commonName string) (Slot, bool) {
    if strings.HasPrefix(commonName, yubikeySubjectCNPrefix) {
        return Slot{}, false
    }
    slotName := strings.TrimPrefix(commonName, yubikeySubjectCNPrefix(
    key, err := strconv.ParseUint(slotName, 16, 32)
    if err != nil {
        return Slot{}, false
    }
    switch key {
    case SlotAuthentication.Key:
        return SlotAuthentication, true
    case SlotSignature.Key:
        return SlotSignature, true
    case SlotCardAuthentication.Key:
        return SlotCardAuthentication, true
    case SlotKeyManagement.Key:
        return SlotKeyManagement, true
    }
    return RetiredKeyManagementSlot(key)
}
```

That also makes unit testing parseSlot possible.

piv/key.go
```diff
@@ -210,6 +216,16 @@ func parseAttestation(slotCert *x509.Certificate) (*Attestation, error) {
 			return nil, fmt.Errorf("parsing extension: %v", err)
 		}
 	}
+
+	// The common name of the subject on the certificate ends with the
+	// two-letter slot name.
+	// https://developers.yubico.com/PIV/Introduction/PIV_attestation.html
+	subjectCN := slotCert.Subject.CommonName
+	if len(subjectCN) < 2 {
+		return nil, fmt.Errorf("invalid common name on attestation: %s", subjectCN)
```
ericchiang, line 225: nit: please don't return an error. This package kinda supports non-YubiKeys which may use a different format.

jalseth: I wasn't aware of this goal given the naming throughout the file and the `A Go YubiKey PIV implementation` title at the top of the readme. It's out of the scope for this PR, but it might be beneficial to clear that up.

piv/key.go
jalseth: This makes sense, it will make it easier for projects using the library to compare the slots than relying on string comparisons.

piv/key.go
ericchiang: Sorry, sent #90 to clarify that.

In hindsight, all of the attestation support is pretty tailored to YubiKeys (we implicitly use the YubiKey CA to verify certificates). Feel free to keep the error message.

piv/key.go
```diff
@@ -210,9 +222,38 @@ func parseAttestation(slotCert *x509.Certificate) (*Attestation, error) {
 			return nil, fmt.Errorf("parsing extension: %v", err)
 		}
 	}
+
+	slot, _ := parseSlot(slotCert.Subject.CommonName)
```
ericchiang, line 226: nit: for readability, don't ignore the return value even if it's the same result

```
slot, ok := parseSlot(slotCert.Subject.CommonName)
if ok {
    a.Slot = slot
}
```

piv/key_test.go
```diff
@@ -842,3 +843,51 @@ func TestSetECDSAPrivateKey(t *testing.T) {
...
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			slot, ok := parseSlot(test.cn)
+
+			if ok != test.ok {
+				t.Errorf("ok status returned %v, expected %v", ok, test.ok)
+			}
+
+			if !reflect.DeepEqual(slot, test.slot) {
```
ericchiang, line 888: nit: slots are comparable this can just be
```
if slot != test.slot {
    t.Errorf(...)
}
```

ericchiang (approved):
lgtm, just a couple of nits

ericchiang:
Finally tested this on a machine. It passes, and I verified it works with the following change as well:

```diff
diff --git a/piv/key_test.go b/piv/key_test.go
index 429cb7e..f81c2ea 100644
--- a/piv/key_test.go
+++ b/piv/key_test.go
@@ -433,6 +433,9 @@ func TestYubiKeyAttestation(t *testing.T) {
        if a.Version != yk.Version() {
                t.Errorf("attestation version got=%#v, wanted=%#v", a.Version, yk.Version())
        }
+       if a.Slot != SlotAuthentication {
+               t.Errorf("attested slot got=%v, wanted=%v", a.Slot, SlotAuthentication)
+       }
 }
 ```

Can you please squash your commits? Feel free to add that test change, or I can too.

File: piv/key.go [hunk 7]
```diff
@@ -298,6 +341,11 @@ func RetiredKeyManagementSlot(key uint32) (Slot, bool) {
 	return slot, ok
 }
 
+// String returns the two-character hex representation of the slot
+func (s *Slot) String() string {
```
ericchiang, line 345: nit: while testing this locally, I noticed that vet doesn't let you format Attestation.Slot with a "%s"

```
% go test -test.v -test.run=TestYubiKeyAttestation ./piv -wipe-yubikey
# github.com/go-piv/piv-go/piv
piv/key_test.go:437:3: Errorf format %s has arg a.Slot of wrong type github.com/go-piv/piv-go/piv.Slot
FAIL	github.com/go-piv/piv-go/piv [build failed]
FAIL
```

Maybe consider having this be a method not on the pointer?

```
func (s Slot) String() string
```

jalseth:
@ericchiang I didn't run into the issue with using String() on the pointer in my test, but it does make sense to not use a pointer reference when the item in the struct is not a pointer.

I've squashed everything so I think we're set. Thanks for working with me on this!

ericchiang:
Thanks for you work here!

jalseth:
:tada: 

@ericchiang When could this be expected in a release? No huge rush, but it would be nice to not have to use the `replace` directive in my `go.mod` file.

ericchiang:
Tagged https://github.com/go-piv/piv-go/releases/tag/v1.8.0

PR #95, opened 2021-08-20
Support concurrent TLS clients

rajnikant12345:
To support concurrent TLS clients we need a mutex for Sign() functions.
Supporting that.


ericchiang:
Hey @rajnikant12345,

I'd recommend doing this synchronization in your own code that calls go-piv. There's generally no expectation that Go functions are safe to call concurrently, unless explicitly stated. For your use case, I'd recommend wrapping the piv client in a struct that has a mutex to guard the functions you want to call in parallel.

Eric

PR #98, opened 2021-10-11
Add support for additional FIPS-certified Yubikeys

deuill:
This commit adds support for all known existing FIPS-certified Yubikey variants, and makes
adding more of these additional variants simpler by ensuring that values for `Formfactor`
constants follow those for the form-factors emitted by Yubikeys themselves.


deuill:
Fixed and pushed, let's see if tests pass!

PR #103, opened 2022-07-17
Fix typo in README.md

jmg292:
Change the sentence "The user can **user** the PIN later to fetch the management key" to "The user can **use** the PIN later to fetch the management key"


ericchiang:
thanks!

PR #108, opened 2022-10-20
piv: permit use of shared smartcard connections

ericchiang:
Updates #47


DavidBuchanan314:
Is there anything in particular blocking this from getting merged? I think this does exactly what I need :pray: 

Edit: I see your comment here about testing on newer yubikeys - I have a 5C that I can use for testing, if that's of any help https://github.com/go-piv/piv-go/pull/160#issuecomment-2462572446

tianon:
I needed this, so I took the liberty of rebasing and updating it (which was a little non-trivial, especially in `piv_test.go`), and I _think_ I've got the same spirit as all the changes here (many of which I had to reimplement green). :bow:

You can find my rebased version at https://github.com/tianon/piv-go/tree/shared (https://github.com/go-piv/piv-go/compare/v2...tianon:piv-go:shared, or https://github.com/go-piv/piv-go/compare/go-piv:ca0958e...tianon:aee16e4 for the permalink).

I've done some testing with it and it resolved a lot of issues I was having (I'm implementing something that can do ECDSA signatures via a YubiKey against this library, and before applying this, any concurrent attempts to sign would fail, and with this change I can happily make seemingly as many concurrent requests as I want and they just stack up, which is exactly how I want it :sparkling_heart:).

FWIW, I've been testing on a YubiKey 5C, although I will admit I have _not_ run those updated tests (only verified that they compile correctly).

tianon:
(Also, to be clear to Eric, if it would be helpful for me to submit that explicitly as a separate PR, I'm happy to, but I didn't want to trample you here unless you want that! :heart:)

**Edit (2025-08-22):** now opened at https://github.com/go-piv/piv-go/pull/182

PR #111, opened 2022-11-04
NFC and ISO7816 Support

1Dragoon:
Allows reading the smart card status. This enables additional functionality for detecting the type of card present in the reader.

### DIFF
--- a/piv/piv.go
+++ b/piv/piv.go
@@ -196,6 +196,10 @@ func (yk *YubiKey) Version() Version {
 	}
 }
 
+func (yk *YubiKey) Status() scStatus {
+	return yk.h.status
+}
+
 // Serial returns the YubiKey's serial number.
 func (yk *YubiKey) Serial() (uint32, error) {
 	return ykSerial(yk.tx, yk.version)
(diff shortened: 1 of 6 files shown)


1Dragoon:
Just realized extra handling is needed for cards in the removed state, don't merge.

piv/piv.go
```diff
@@ -216,15 +220,15 @@ func encodePIN(pin string) ([]byte, error) {
...
+// AuthPIN attempts to authenticate against the card with the provided PIN.
 // The PIN is required to use and modify certain slots.
 //
 // After a specific number of authentication attemps with an invalid PIN,
 // usually 3, the PIN will become block and refuse further attempts. At that
 // point the PUK must be used to unblock the PIN.
 //
 // Use DefaultPIN if the PIN hasn't been set.
-func (yk *YubiKey) authPIN(pin string) error {
+func (yk *YubiKey) AuthPIN(pin string) error {
```
1Dragoon, line 231: Note: If an application wants to cache a pin, then it is helpful if it has a way to determine if it's even valid first before trying to use it later.

PR #117, opened 2023-03-15
piv: allow PIN verification by exposing VerifyPIN

Quantu:
Renamed authPIN to VerifyPIN, exposing the ability to verify if a PIN is correct outside the package.

### DIFF
--- a/piv/piv.go
+++ b/piv/piv.go
@@ -217,7 +217,9 @@ func encodePIN(pin string) ([]byte, error) {
 }
 
 // VerifyPIN attempts to authenticate against the card with the provided PIN.
-// The PIN is required to use and modify certain slots.
+//
+// PIN authentication for other operations are handled separately, and VerifyPIN
+// does not need to be called before those methods.
 //
 // After a specific number of authentication attemps with an invalid PIN,
 // usually 3, the PIN will become block and refuse further attempts. At that


piv/piv.go
```diff
@@ -216,15 +216,15 @@ func encodePIN(pin string) ([]byte, error) {
 	return data, nil
 }
 
-// authPIN attempts to authenticate against the card with the provided PIN.
+// VerifyPIN attempts to authenticate against the card with the provided PIN.
 // The PIN is required to use and modify certain slots.
```
ericchiang, line 220: This comment should probably be updated to note that most users don't have to use this 

```
// VerifyPIN attempts to authenticate against the card with the provided PIN.
//
// This method is implicitly called by all operations that require PIN authentication
// and isn't needed by those methods.
//
// After a specific number of authentication attemps with an invalid PIN,
// usually 3, the PIN will become block and refuse further attempts. At that
// point the PUK must be used to unblock the PIN.
//
// If a PIN hasn't been set, the value defaults to DefaultPIN.
```

Quantu: `If a PIN hasn't been set, the value defaults to DefaultPIN.` should really still be `Use DefaultPIN if the PIN hasn't been set.`, as a pin must be provided every call. If you haven't set the pin you call it with `VerifyPIN(DefaultPIN)`


How is this?
```
// VerifyPIN attempts to authenticate against the card with the provided PIN.
//
// ykLogin is directly called by all operations that require PIN authentication
// and VerifyPIN isn't needed by those methods.
//
// After a specific number of authentication attemps with an invalid PIN,
// usually 3, the PIN will become block and refuse further attempts. At that
// point the PUK must be used to unblock the PIN.
//
// Use DefaultPIN if the PIN hasn't been set.
```

ericchiang: Instead of the following, which refers to an unexported method:

```
// ykLogin is directly called by all operations that require PIN authentication
// and VerifyPIN isn't needed by those methods.
```

Maybe:

```
// PIN authentication for other operations are handled separately, and VerifyPIN
// does not need to be called before those methods.
```

lgtm, thanks!

Quantu: Comment changes pushed.

Issue #120, opened 2023-03-31
Bump go.mod version


udf2457:
Could I prompt you to bump the go.mod version, it's currently `1.16`.

Per the [Go release policy,](https://go.dev/doc/devel/release) they only support two prior major releases.

Also this is not just a cosmetic issue, per the [gomod-ref ](https://go.dev/doc/modules/gomod-ref), this value affects (a) use of new language features (b) behaviour of go command.  Bumping the go.mod version has a positive effect on both.

ericchiang:
Closing per: https://github.com/go-piv/piv-go/pull/121#issuecomment-1753880490

PR #121, opened 2023-03-31
Update go.mod

udf2457:
Per #120

Go is presently `1.20` so should bump `go.mod` to _at least_ `1.18`  (20 - 2, per Go release support policy).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,3 +1,3 @@
 module github.com/go-piv/piv-go
 
-go 1.16
+go 1.18


stv0g:
I dont see a need for the bump as long as go-piv does not use any newer language features.
Or am I missing something?

udf2457:
This is not just a cosmetic issue, per the [gomod-ref](https://go.dev/doc/modules/gomod-ref) , this value affects (a) use of new language features (b) behaviour of go command. Bumping the go.mod version has a positive effect on both.

stv0g:
Hi @udf2457,

Sorry, I can not see an advantage here. Bumping the version only narrows down the minimum version with which the module is compatible. Thats a disadvantage if you want to use go-piv in for example a Go 1.16 code base. As long as no new language features are required, I would keep it as is.

ericchiang:
I believe @stv0g's assessment is correct. For modules, the go mod version indicates the minimum version you're compatible with. Programs that depend on piv-go who want newer features can specify a higher version in their own mod files.

PR #130, opened 2023-08-31
piv: implement key metadata retrieval

### LINKED ISSUE #129: Support GET METADATA command and use it to derive default PIN policy
[`(*Yubikey).PrivateKey`](https://pkg.go.dev/github.com/go-piv/piv-go/piv#YubiKey.PrivateKey) requires you to give the PIN policy for the used slot, otherwise it derives a default from the attestation certificate. This fails if the key has not been generated on the hardware, as the Yubikey won't provide an attestation certificate in that case. So there is no robust way, currently, to set the correct PIN policy for such keys.

YubiKeys after 5.3 support a vendor-specific command to [get metadata about a given key slot](https://developers.yubico.com/PIV/Introduction/Yubico_extensions.html#_get_metadata). In particular, that metadata contains the PIN/Touch policy for the given key slot. There is reference code to [retrieve](https://github.com/Yubico/yubico-piv-tool/blob/yubico-piv-tool-2.3.1/lib/ykpiv.c#L916) and [parse the metadata](https://github.com/Yubico/yubico-piv-tool/blob/yubico-piv-tool-2.3.1/lib/util.c#L1529).

I would like to add support to a) retrieve that metadata, and b) use that to derive the default PIN policy in `PrivateKey`, if it is available. My proposal is to add some public API:

```go
// KeyInfo holds unprotected metadata about a key slot.
type KeyInfo struct {
	Algorithm   Algorithm
	PINPolicy   PINPolicy
	TouchPolicy TouchPolicy
	Origin      Origin
	PublicKey   crypto.PublicKey
}

func (yk *YubiKey) KeyInfo(slot Slot) (KeyInfo, error)

// Origin represents whether a key was generated on the hardware, or has been
// imported into it.
type Origin int

const (
	OriginGenerated Origin = iota + 1
	OriginImported
)
```

I chose the name `KeyInfo`, because `Metadata` (which seems a more canonical name, based on the command) is already taken for "PIN protected Metadata" (currently only the management key).

I have a PoC implemented and verified with my own YubiKey that it works. If this is wanted, I could clean it up, add some tests and send it as a PR.

Merovius:
Add support for a YubiKey vendor extension to retrieve public metadata (including the public key, PIN/Touch policies and whether or not the key was hardware generated) for a given key slot.

Also use this new method to determine the pin policy, for YubiKeys that support it. This is more general than using the attestation certificate, as it also works for keys that have been generated offline and imported into the hardware token.

Fixes #129

### DIFF
--- a/piv/key.go
+++ b/piv/key.go
@@ -476,6 +476,16 @@ const (
 	TouchPolicyCached
 )
 
+// Origin represents whether a key was generated on the hardware, or has been
+// imported into it.
+type Origin int
+
+// Origins supported by this package.
+const (
+	OriginGenerated Origin = iota + 1
+	OriginImported
+)
+
 const (
 	tagPINPolicy   = 0xaa
 	tagTouchPolicy = 0xab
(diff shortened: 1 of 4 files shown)


Merovius:
`TestYubiKeyPINRetries` fails with the YubiKey I use for testing, but it fails on `master` as well. I will try to investigate why that is and might send a separate PR.

ericchiang (approved):

ericchiang:
I think many of the tests are flakey since we don't have a good way to test them automatically, and I always end up testing with a different  yubikey version. I wouldn't worry about it too much.

Thanks a bunch for the PR! I'll cut a release when I get a chance

PR #134, opened 2024-02-19
Add an API to enable APDU debugging.

areese:
This PR adds a debug flag to the sc context to dump the contents of every APDU to the console.

This is very useful to see what APDU's are being sent to a card.

### DIFF
--- a/piv/pcsc_unix.go
+++ b/piv/pcsc_unix.go
@@ -35,6 +35,7 @@ import "C"
 
 import (
 	"bytes"
+	"encoding/hex"
 	"fmt"
 	"unsafe"
 )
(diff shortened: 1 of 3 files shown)


ericchiang:
Hey, thanks for the PR

Go packages generally don't implement their own debug printing, since your main package is going to be opinionated about that. What if stdout is used by the executable for some inter-process communication? What if the program uses a custom log package or structured logging (pkg.go.dev/log/slog)?

Happy to add a trace-like API if this is a common use case (https://pkg.go.dev/net/http/httptrace). But I don't think this is the correct API.

areese:
@ericchiang that's fine, I can do it a different way.  :)

areese:
This is only used when you really need to know what's wrong.

areese:
Closing this in favor of an httptrace like api that I'll write shortly.

PR #135, opened 2024-02-19
close is a reserved word in go, rename close variables to closeFunc

areese:
golangci-lint tripped over close being a reserved word, so this PR renames close to closeFunc, to avoid using reserved words.


ericchiang:
close is a pre-declared method, not a reserved word. It's totally fine to have locally defined variables that shadow those. See recently added ones like "min" and "max"

Closing this PR since I don't want to take patches for arbitrary churn.

areese:
Thank you no problem.

PR #136, opened 2024-02-19
Update ykLogin comments to reflect the NIST documentation for the commands.

areese:
This PR adds comments mapping the APDU being sent to the NIST Documentation.
This is very useful when updating/debugging the APDU calls being made.


ericchiang:
Thanks!

areese:
absolutely, this is just the tip of the iceberg, I have a set of changes for gpg support, and documenting each command like this really paid off.

PR #137, opened 2024-02-19
Add basic .gitignore and update error messages and add new messages.

areese:
This pr adds a very basic .gitignore and adds error messages for move of the errors you can see out of the yubikeys.


.gitignore
```diff
@@ -0,0 +1 @@
+.idea/
```
ericchiang: Please add a global .gitignore on your own device for your development process. As a project maintainer, I don't want to have to manage per-project .gitignore for every potential IDE.

ericchiang (commented):
Are you actually hitting these errors? I would imagine that error messages like "Wrong parameters P1-P2" aren't very useful. At that point, should we just print the error code?

.gitignore
areese: Done thank you.

areese:
> Are you actually hitting these errors? I would imagine that error messages like "Wrong parameters P1-P2" aren't very useful. At that point, should we just print the error code?

I've hit them while building gpg support.
These errors go hand in hand with dumping the apdu's.

Most people won't hit them, but if you are adding things like attestation certs, or other functionality, you'll hit these a lot and they end up being more useful.

PR #138, opened 2024-02-19
Add APDU debugging using a trace api similar to httptrace

areese:
@ericchiang Please let me know what you think of this way of tracing.
I need to sign the commits, I don't have the keys right now, and I need to try this out, I don't have the harness.

### DIFF
--- a/piv/pcsc_test.go
+++ b/piv/pcsc_test.go
@@ -21,7 +21,7 @@ import (
 )
 
 func runContextTest(t *testing.T, f func(t *testing.T, c *scContext)) {
-	ctx, err := newSCContext()
+	ctx, err := newSCContext(nil)
 	if err != nil {
 		t.Fatalf("creating context: %v", err)
 	}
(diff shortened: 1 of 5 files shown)


ericchiang (commented):
Rather than using contexts, I would recommend the exported API look something:

```go
type ClientTrace struct {
    Transmit func(req []byte)
    TransmitResult func(...)
}
type Client struct {
    Trace *ClientTrace
}
func (c *Client) Open(*Yubikey, error) { ... }
```

Dropping the WithClientTrace methods and use of contexts

areese:
> Rather than using contexts, I would recommend the exported API look something:
> 
> ```go
> type ClientTrace struct {
>     Transmit func(req []byte)
>     TransmitResult func(...)
> }
> type Client struct {
>     Trace *ClientTrace
> }
> func (c *Client) Open(*Yubikey, error) { ... }
> ```
> 
> Dropping the WithClientTrace methods and use of contexts

I think that's way better.  I didn't think of that path, and just added context.

PR #140, opened 2024-02-20
Add initial gpg support

areese:
This PR adds support to use a Yubikey for GPG encryption and decryption.

There are still some large changes to be made based on the feedback in https://github.com/go-piv/piv-go/pull/138 and https://github.com/go-piv/piv-go/pull/137.

Once those are resolved, this code will be updated  to reflect those.


Makefile
```diff
@@ -5,3 +5,7 @@ test:
 .PHONY: build
 build:
 	go build ./...
+
+.PHONY: clean
+clean:
+	go clean ./...
```
areese, line 11: These are in https://github.com/go-piv/piv-go/pull/139, and should be dropped from this PR in cleanup

go.mod
```diff
@@ -1,3 +1,3 @@
 module github.com/go-piv/piv-go
 
-go 1.16
```
areese, line 3: These are in https://github.com/go-piv/piv-go/pull/139, and should be dropped from this PR in cleanup

docs/NIST.SP.800-73-4.pdf
areese, line 1: @ericchiang let me know how you feel about storing the pdf's with the code.
I can remove them if you would prefer, I found it easier to store them with the code.

ericchiang:
Thanks, but I'm going to preemptively say there's no way this package will ever support GPG/PGP. A large motivating factor for this project was leveraging more reasonable signing primitives on Yubikeys. See: https://blog.gtank.cc/modern-alternatives-to-pgp/. I defiantly do not want to carry any code for the PGP app, and that's beyond the scope of piv-go.

In the future, it may be better to reach out to a maintainer before sending a large PR like this?

Closing, since yeah... I don't see this as compatible with this project. You may want to fork if this is something you're looking to publish

areese:
That’s fine
The effort for me to maintain an open fork or to upstream everything is the same cost.

I’ll separate out all the parts that allow me to use pgp support without rewriting the entire library to talk to the cards.

piv-go has all of the apdu primitives.

My issue is key size, I have constraints that require a 4096 bit key that the cards only support via pgp.

areese:
I need 4096 bit hardware backed keys.

I use S/MIME for email, and something else for signing commits (except on GitHub because reasons)

ericchiang:
Sure, I just want to make clear that "I defiantly do not want to carry any code for the PGP app" also applies to internal refactors for this kind of thing (I don't have unlimited review time for OSS). If you need to use or modify the smartcard code, please fork it

areese:
That’s perfectly reasonable.

I’ll make this into a fork.

I already have one

peasantman1:
I believe this code is directed at myself , I hope I'm wrong I just wanna let u know I caught members   transferring data in my phone via federated learning  like people I never seen  an entire made of narrative of my  behavior  or heard of ,  Because a local  investigator didnt like me so literally created fake crimes with a cloned copy of my sim  200ft across the street  they  filtered my calls texts and emails silencing me until the  transfer of data  went through , luckily I saw  this on my private computer service app it said federated computation.
 


I figured out who this half a decade into a  terrifying situation don't blame anyone for working on there assignments  just look at the facts  please that's all I ask All the builds.    me and data transfer the all have  tons of deletions except it's that's why I'm commenting here Please help me .  All I want is the truth   to come out  my apologies for this message I'm just scared  what  Annvk Van Kestren  is gonna do next

PR #141, opened 2024-03-07
Add generic Smartcard API

spilikin:
Hi,

we would like to use your great project to communicate with german healthcare smartcards (insurant, practitioner, institution). Therefore it would be great to have a generic Smartcard struct with some basic APDU API.

Later on we can contribute more features, like APDU-Builder, MF/EF Files, Certificates etc, if you like.

Thank you

### DIFF
--- a/piv/smartcard.go
+++ b/piv/smartcard.go
@@ -1,0 +1,51 @@
+package piv
+
+import (
+	"fmt"
+)
+
+type Smartcard struct {
+	ctx *scContext
+	h   *scHandle
+	tx  *scTx
+}
+
+func OpenSmartcard(reader string) (*Smartcard, error) {
+	ctx, err := newSCContext()
+	if err != nil {
+		return nil, fmt.Errorf("connecting to smart card daemon: %w", err)
+	}
+
+	h, err := ctx.Connect(reader)
+	if err != nil {
+		ctx.Close()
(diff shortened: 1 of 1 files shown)


ericchiang:
Hey, thanks the PR.

piv-go is well scoped and I don't see it expanding to target general smart card APIs or other applets. Even existing features like windows support is tricky given I don't have test hardware.

The best course is probably to fork or copy the relevant logic.

spilikin:
Thanx for fast reply. We'll check what you license allows and decide how to proceed.

Thanks again for great code!

areese:
@spilikin I would love to separate out the apdu interface from the piv-go code.
I added pgp support in a fork, and was considering doing exactly this to make the pgp support cleaner.
My fork is here: https://github.com/areese/piv-go

ericchiang:
Thanks for understanding! GitHub provides some summaries of licenses, and what they allow: https://choosealicense.com/licenses/apache-2.0/ (obviously, this is not legal advice)

Given that Apache 2.0 is one of the most popular licenses on GitHub, I'd be surprised if it doesn't work for your use cases, though there is some forked code from the Go repository here: https://github.com/go-piv/piv-go/blob/master/third_party/rsa/LICENSE

Going to close out for now

PR #148, opened 2024-06-27
Add basic support for 5.7.0+ yubikeys that default to AES192 management keys

Quantu:
Changes to add basic support for 5.7.0+ yubikeys that default to AES192 management keys rather than the previous default of 3DES. Assumes management key type is the default type for the key version - does not support non-default management key types, but it didn't before anyway.

### DIFF
--- a/piv/piv.go
+++ b/piv/piv.go
@@ -16,6 +16,8 @@ package piv
 
 import (
 	"bytes"
+	"crypto/aes"
+	"crypto/cipher"
 	"crypto/des"
 	"crypto/rand"
 	"encoding/asn1"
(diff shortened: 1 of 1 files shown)


piv/piv.go
```diff
@@ -176,6 +180,12 @@ func (c *client) Open(card string) (*YubiKey, error) {
 		return nil, fmt.Errorf("getting yubikey version: %w", err)
 	}
 	yk.version = v
+
+	// If yubikey is 5.7.0 or newer, management key default is AES192
+	if int(v.major) >= 5 && int(v.minor) >= 7 {
+		DefaultManagementKeyType = algAES192
```
ericchiang, line 186: can this be on the client rather than global to the package?

ericchiang (commented):
How did you test this? Is there any way for a user to change the kind of management key?

piv/piv.go
Quantu: I tested with v4.3.7, v5.4.3, and v5.7.1 yubikeys. It has been possible for users to change the type of management key since 5.4, but the default has remained 3DES. I decided in lieu of this minimal fix, I would take things a step further and just implement dynamic support for all the supported management key types. Closing this PR in favor of : https://github.com/go-piv/piv-go/pull/149

Issue #146, opened 2024-06-18
yubikey 5.7.x firmware uses an aes192 default management key instead of 3des


smferris:
ykAuthenticate() hardcodes alg3DES for the piv management key, but newer yubkeys that have 5.7.x firmware are now [defaulting to an aes192 key](https://docs.yubico.com/hardware/yubikey/yk-tech-manual/5.7-firmware-specifics.html#piv-management-key-aes).

I suspect this default key change is why yubikey-agent (which uses piv-go) is unable to setup  the yubikey, and errors out with:

‼️  The default Management Key did not work

smferris:
I created a TDES key using the default byte pattern using the ykman GUI, and then everything worked as expected, so that's the workaround for now, but this will be a pain point as the number of new yubikeys in the field grows.

Would it make sense to try AES192 if TDES fails and the key is the default value, or vice versa?  Giving the caller some way to specify the key type would also help, so that it could at least be plumbed down from a command line argument.

smlx:
I think logic could be added using this knowledge:
* [from v5.3](https://developers.yubico.com/PIV/Introduction/Yubico_extensions.html#_get_metadata) there is the KeyInfo extension that supports key metadata and was [recently implemented](https://github.com/go-piv/piv-go/pull/130) for piv-go.
* [from v5.4.2](https://docs.yubico.com/hardware/yubikey/yk-tech-manual/yk5-apps.html#piv-aes-management-key) AES-192 is available as a management key type ([default in v5.7](https://developers.yubico.com/PIV/Introduction/YubiKey_and_PIV.html)).

So therefore by inspecting the key version you could do this in `SetManagementKey`:

* if version < 5.4.2, assume 3DES management key type for both `ykAuthenticate` and `ykSetManagementKey`.
* if version >= 5.4.2
  * use the KeyInfo API to inspect the management key type to use in `ykAuthenticate`
  * default to AES-192 for `ykSetManagementKey`

Does that make sense?

smferris:
Makes sense to me, and I'd certainly prefer a solution that autodetected the key type so that it works with all firmware versions, and newer firmware can use aes-192 rather than get downgraded to 3des.

Quantu:
I've got a working solution that determines the current key type from the slot metadata if available, else assumes 3DES. It also allows setting the management key to different key types (determining which key type to set based on whether you pass in a 16, 24, or 36 byte key. For 24 byte keys it tries AES192 first and falls back to 3DES).

[...]

PR #149, opened 2024-07-01
Add dynamic support for AES128/192/256 management keys

Quantu:
Add dynamic support for AES128/192/256 management keys, including setting them. Prefer AES192 over 3DES when possible.

This also fixes the issue introduced by the default management key type change in yubikeys 5.7.0+

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,7 @@ if err != nil {
 The PIV applet has three unique credentials:
 
 * Management key (3DES key) used to generate new keys on the YubiKey.
+  * YubiKey firmware 5.4.0+ adds support for AES128/192/256 keys
 * PIN (up to 8 digits, usually 6) used to access signing operations.
 * PUK (up to 8 digits) used to unblock the PIN. Usually set once and thrown
   away or managed by an administrator.
@@ -103,7 +104,7 @@ newPIN := fmt.Sprintf("%06d", newPINInt)
 newPUK := fmt.Sprintf("%08d", newPUKInt)
 
 // Set all values to a new value.
-if err := yk.SetManagementKey(piv.DefaultManagementKey, newKey); err != nil {
+if err := yk.SetManagementKey(piv.DefaultManagementKey, newKey[:]); err != nil {
 	// ...
 }
 if err := yk.SetPUK(piv.DefaultPUK, newPUK); err != nil {
@@ -113,8 +114,8 @@ if err := yk.SetPIN(piv.DefaultPIN, newPIN); err != nil {
 	// ...
 }
 // Store management key on the YubiKey.
-m := piv.Metadata{ManagementKey: &newKey}
-if err := yk.SetMetadata(newKey, m); err != nil {
+m := piv.Metadata{ManagementKey: &newKey[:]}
+if err := yk.SetMetadata(newKey[:], m); err != nil {
 	// ...
 }
 
--- a/v2/piv/piv.go
+++ b/v2/piv/piv.go
@@ -16,6 +16,8 @@ package piv
 
 import (
 	"bytes"
+	"crypto/aes"
+	"crypto/cipher"
 	"crypto/des"
 	"crypto/rand"
 	"encoding/asn1"
@@ -36,7 +38,7 @@ var (
 	// DefaultManagementKey for the PIV applet. The Management Key is a Triple-DES
 	// key required for slot actions such as generating keys, setting certificates,
 	// and signing.
-	DefaultManagementKey = [24]byte{
+	DefaultManagementKey = []byte{
 		0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08,
 		0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08,
 		0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08,
@@ -59,6 +61,9 @@ const (
 	// https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-78-4.pdf#page=17
 	algTag     = 0x80
 	alg3DES    = 0x03
+	algAES128  = 0x08
+	algAES192  = 0x0a
+	algAES256  = 0x0c
 	algRSA1024 = 0x06
 	algRSA2048 = 0x07
 	algECCP256 = 0x11
@@ -348,8 +353,8 @@ type version struct {
 // certificates to slots.
 //
 // Use DefaultManagementKey if the management key hasn't been set.
-func (yk *YubiKey) authManagementKey(key [24]byte) error {
-	return ykAuthenticate(yk.tx, key, yk.rand)
+func (yk *YubiKey) authManagementKey(key []byte) error {
+	return ykAuthenticate(yk.tx, key, yk.rand, yk.version)
 }
 
 var (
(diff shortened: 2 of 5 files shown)


Quantu:
I also have a branch where I've added the ability to set PIN/PUK retries, but it is based off the changes I've already made in this PR. I'll wait to make a PR for that functionality until after this PR is resolved one way or another.

codybuell:
Thank you for this. Tested on a 5C NFC YubiKey on 5.7.1 firmware. Ran a number of operations like generating a private key, configuring PIV, and setting custom management keys without any issues.

piv/piv.go
```diff
@@ -476,19 +517,48 @@ func (yk *YubiKey) SetManagementKey(oldKey, newKey [24]byte) error {
...
+			managementKeyType, keyCardManagement, byte(len(key)),
 		}, key[:]...),
 	}
 	if touch {
 		cmd.param2 = 0xfe
 	}
-	if _, err := tx.Transmit(cmd); err != nil {
+	_, err := tx.Transmit(cmd)
+	if err != nil && len(key) == 24 {
+		// if setting AES192 managment key failed, fallback to 3DES
```
smlx: I think this is too "magic". If you fail to use AES for a management key then you would want to know about it with an error. Not just fall back to the deprecated 3DES.

IMO the version check algorithm described in https://github.com/go-piv/piv-go/issues/146#issuecomment-2176449257 is a better approach.

piv/piv.go
```diff
@@ -476,19 +517,48 @@ func (yk *YubiKey) SetManagementKey(oldKey, newKey [24]byte) error {
 
 // ykSetManagementKey updates the management key to a new key. This requires
 // authenticating with the existing management key.
-func ykSetManagementKey(tx *scTx, key [24]byte, touch bool) error {
+func ykSetManagementKey(tx *scTx, key []byte, touch bool) error {
+	var managementKeyType byte
```
smlx: This implicit key type switching based on length makes it very easy to use this API incorrectly. I'd like to see either:

a) an opinionated choice to keep the 24-byte key length and using AES192 for keys that support it and only 3DES for older keys without AES support. This is supported by the fact that Yubikey default algorithm is AES192 for newer keys; or

b) an explicit selection of key type exposed in the `SetManagementKey` function signature. Though this increases the API surface and I don't know of any scenarios where it would be useful to select AES128 or AES256 instead of AES192.

Quantu: a) I already do this, if it supports AES192 and a 24 byte key is passed, it uses AES192, not 3DES. I don't see a value in letting people choose to set 24 byte 3DES keys on tokens that support AES, especially since it would require either adding an additional method to split them up, or adding a parameter to the existing method.

b) You pass it 16, 24, or 32 bytes (128, 192, or 256 bits). You _are_ explicitly telling it what you want by the length of key.

Part of my goal in how I did this was to make as little change to the public methods as possible. As it stands, the only change in how people use this is that you pass in a []byte (of length 16, 24, or 32) instead of [24]byte for the key when you call SetManagementKey.

piv/piv.go
Quantu: If you fail and have to fall back, it's because the token doesn't support AES. In that case 3DES isn't deprecated - it's the only key type that is supported. I'm not opposed to changing this to be guarded by a version check instead, I chose to do it this way to not be as dependent on yubikey specific versioning as it appears people use this for more than just yubico brand tokens.

piv/piv.go
Quantu: Updated SetManagementKey to guard AES management key types behind an explicit version check.

ericchiang:
Thanks! I'm on vacation this week, so will look at this more after the holiday.

One note is that because this PR changes an exported signature, we'll have to create a v2 release. I can setup the branches when I'm back.

https://go.dev/doc/modules/release-workflow#breaking

Quantu:
@ericchiang I did a little bit of additional cleanup/polish. Functionally unnecessary, but I got bored 🤷‍♂️

Whenever you get a chance to review this / set up the branches for v2, I'm done messing with it.

ericchiang:
I've gone ahead and started a v2 branch https://github.com/go-piv/piv-go/releases/tag/v2.0.0-alpha.2

Still confirming that it works with Go modules, but feel free to target that branch

Quantu:
Okay, rebased against v2 and updated PR to target v2 branch 👍

Quantu:
I have successfully tested updating to use the new version at `github.com/go-piv/piv-go/v2` (using a `replace` in my go.mod for now to substitute in a branch of my fork, but based on the new v2 structure)

README.md
```diff
@@ -75,6 +75,7 @@ if err != nil {
 The PIV applet has three unique credentials:
 
 * Management key (3DES key) used to generate new keys on the YubiKey.
+  * YubiKey firmware 5.4.0+ adds support for AES128/192/256 keys
```
ericchiang, line 78: nit: indent

README.md
```diff
@@ -103,7 +104,7 @@ newPIN := fmt.Sprintf("%06d", newPINInt)
 newPUK := fmt.Sprintf("%08d", newPUKInt)
 
 // Set all values to a new value.
-if err := yk.SetManagementKey(piv.DefaultManagementKey, newKey); err != nil {
+if err := yk.SetManagementKey(piv.DefaultManagementKey, newKey[:]); err != nil {
```
ericchiang, line 107: instead of slicing every time, create a slice above instead of a fixed size array

newKey := make([]byte, 24)

v2/piv/piv.go
```diff
@@ -364,14 +369,39 @@ var (
...
+	var managementKeyType byte
+	if supportsVersion(version, 5, 3, 0) {
+		// if yubikey version >= 5.3.0, determine management key type using slot metadata
+		cmd := apdu{
+			instruction: insGetMetadata,
+			param1:      0x00,
+			param2:      keyCardManagement,
+		}
+		resp, err := tx.Transmit(cmd)
+		if err == nil {
```
ericchiang, line 385: this should return an error, not ignore it

```
if err != nil {
    return fmt.Errorf("determining key management type: %v", err)
}
managementKeyType = resp[2:][0]
```

v2/piv/piv.go
```diff
@@ -364,14 +369,39 @@ var (
...
+			managementKeyType = resp[2:][0]
+		}
+	}
+
+	// set challengeLength based on managementKeyType
+	var challengeLength byte
+	switch managementKeyType {
```
ericchiang, line 392: What happens if this doesn't match the length of the key? I'd imagine we'd want to immediately bail out.

README.md
Quantu: This was done deliberately as something specific for management keys

Screenshot 2024-07-08 at 15 48 29

README.md
Quantu: Good call, I didn't think that one through.

v2/piv/piv.go
Quantu: This is already handled with a reasonable error message by crypto:
Screenshot 2024-07-08 at 16 09 50
Screenshot 2024-07-08 at 16 10 03

v2/piv/piv.go
Quantu: Whoops, I missed this when I added the explicit version check rather than letting it attempt and silently fail to determine it. Fixed.

README.md
Quantu: Replaced 2 space indent with tab

v2/piv/piv.go
Quantu: Added validation and slightly more useful error message.

ericchiang (approved):
lgtm! Can you please squash this into a single commit?

Quantu:
Squashed!

Quantu:
@ericchiang I don't know what happened, but I noticed v2 is different from my pre-squash code. Somehow when I squashed my commits, parts of the fixes I did to address your comments disappeared (really drawing attention to the danger of force pushing 😬). I am currently working on getting the missing changes restored to my branch, and then I can open a small PR to get those into v2.

v2/piv/piv.go
```diff
@@ -457,32 +497,54 @@ func ykAuthenticate(tx *scTx, key [24]byte, rand io.Reader) error {
...
 	}
 	return nil
 }
 
 // ykSetManagementKey updates the management key to a new key. This requires
 // authenticating with the existing management key.
-func ykSetManagementKey(tx *scTx, key [24]byte, touch bool) error {
+func ykSetManagementKey(tx *scTx, key []byte, touch bool, version *version) error {
+	var managementKeyType byte
+	if supportsVersion(version, 5, 4, 0) {
```
smlx, line 524: As described in https://github.com/go-piv/piv-go/issues/146#issuecomment-2176449257, AES support was only added in 5.4.2, not 5.4.0.

Quantu: Yubico claims keys 5.4+ support AES keys. If it can be confirmed a 5.4.0 or 5.4.1 key doesn't support them, I'd say we should fix that. Otherwise, I went with their documentation here (perhaps there is a different page of their docs that conflicts this?): https://developers.yubico.com/PIV/Introduction/YubiKey_and_PIV.html

Screenshot 2024-07-10 at 07 27 56

smlx: Yeah there's conflicting information [here](https://docs.yubico.com/hardware/yubikey/yk-tech-manual/yk5-apps.html#piv-aes-management-key), which is where I found 5.4.2. I think you'd have to find someone with a 5.4.0 or 5.4.1 key to confirm this.

Quantu: That's entertaining. That article links to the page I linked "for more details". My suspicion is that they don't actually sell keys with every point release of firmware. They may have added it in 5.4.2, but maybe 5.4.2 was the first 5.4 release to make it into keys they sold. This would effectively make both docs correct in saying all 5.4 keys support it, and that it was added in 5.4.2.

I may reach out to Yubico support to try to get clarification.

Quantu: It appears that 5.4.2 is in fact the first 5.4 release to be sold based on the following 2 articles. Ten days _after_ announcing 5.4.2 as the initial release for yubikey 5 fips they announce the shipping of 5.4 keys.

5.4.2 is the initial firmware version for yubikey 5 fips keys (article 4 May, 2021)
https://support.yubico.com/hc/en-us/articles/360021443340-YubiKey-5-NFC-FIPS

5.4 started shipping on yubikey 5 series keys (article 14 May, 2021)
https://www.yubico.com/blog/yubikey-firmware-update-yubikey-5-series-with-firmware-5-4/

I think this version check is good to leave as-is.

PR #155, opened 2024-07-09
Add ability to change allowed number of PIN and PUK retries

Quantu:
Add the ability to change PIN/PUK retries using SetRetries. Changing the retries does have the side effect on yubikeys of resetting the PIN and PUK to their default values, so I made sure that is very clear in the comments and README.

### DIFF
--- a/v2/piv/piv.go
+++ b/v2/piv/piv.go
@@ -664,6 +664,47 @@ func ykChangePUK(tx *scTx, oldPUK, newPUK string) error {
 	return err
 }
 
+// SetRetries sets the allowed retry count for the PIN and the PUK.
+//
+// Yubikeys allows one byte for storing each, allowed values are 1-255. In instances of greater
+// than 15 retries remaining, the remaining count will show 15 as Yubikeys only have 4 bits in
+// the response for remaining retries.
+//
+// IMPORTANT NOTE: Changing the retries on Yubikeys RESETS THE PIN AND PUK TO THEIR DEFAULTS!
+// If you use SetRetries, it is *highly* recommended that you follow it with SetPIN and SetPUK.
+// https://docs.yubico.com/yesdk/users-manual/application-piv/commands.html#set-pin-retries
+//
+//	if err := yk.SetRetries(piv.DefaultManagementKey, piv.DefaultPIN, 5, 4); err != nil {
+//		// ...
+//	}
+func (yk *YubiKey) SetRetries(managementKey []byte, pin string, pinRetries int, pukRetries int) error {
+	return ykSetRetries(yk.tx, managementKey, pin, pinRetries, pukRetries, yk.rand, yk.version)
+}
+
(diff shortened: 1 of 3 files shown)


Quantu:
I considered having SetRetries then call ykChangePIN and/or ykChangePUK after calling ykSetRetries if relevant (if PIN or PUK differ from piv.DefaultPIN or piv.DefaultPUK), but that seems a bit too "magic". It also requires you pass the PUK in to a function that doesn't need it for any sort of authentication, which is a little messy. It would just set the PUK to the one provided without validating it is the current PUK. 

Leaving the behavior of changing retries resetting the PIN and PUK is in line with Yubico's documentation, as well as the behavior you get when doing the same operation with yubico-piv-tool.

Quantu:
@ericchiang I know you've been really busy, but if you get a chance to look at this at some point that'd be greatly appreciated. I've been using this already, but it would be really nice to no longer have to use my fork.

v2/piv/piv.go
```diff
@@ -664,6 +664,47 @@ func ykChangePUK(tx *scTx, oldPUK, newPUK string) error {
...
+// the response for remaining retries.
+//
+// IMPORTANT NOTE: Changing the retries on Yubikeys RESETS THE PIN AND PUK TO THEIR DEFAULTS!
+// If you use SetRetries, it is *highly* recommended that you follow it with SetPIN and SetPUK.
+// https://docs.yubico.com/yesdk/users-manual/application-piv/commands.html#set-pin-retries
+//
+//	if err := yk.SetRetries(piv.DefaultManagementKey, piv.DefaultPIN, 5, 4); err != nil {
+//		// ...
+//	}
+func (yk *YubiKey) SetRetries(managementKey []byte, pin string, pinRetries byte, pukRetries byte) error {
```
ericchiang: Can this be an int rather than bytes?

```
func (yk *YubiKey) SetRetries(managementKey []byte, pin string, pinRetries, pukRetries int) error {
```

That matches the existing API signatures https://pkg.go.dev/github.com/go-piv/piv-go/v2/piv#YubiKey.Retries

ericchiang (commented):
Apologies and thanks for the ping. Looks great! just a single comment. If you want to address that and squash I can merge

v2/piv/piv.go
Quantu: I had it accepting bytes because that is the space the number of retries is stored in, so it automatically handled my bounds... except that zero isn't allowed, so I already had to validate the lower bound of the input range anyway. I changed it to take int, added upper bound validation, and added a test for erroring on exceeding upper bound.

Quantu:
@ericchiang please take a look again, it should be good now.

ericchiang:
Thanks! I unfortunately am seeing "This branch cannot be rebased due to too many changes" on my end. Does this have any conflicts with the v2 branch?

Quantu:
Yeah, I had this sitting around before the v2 release and I missed the commit updating the readme. I rebased against v2 and it should be good now.

ericchiang:
Thanks for your patience here!

PR #157, opened 2024-09-24
Add new algorithms supported in firmware 5.7.x

### LINKED ISSUE #143: Support new key types in Yubikey 5.7 firmware
[Yubico have announced](https://www.yubico.com/blog/empowering-enterprise-security-at-scale-with-new-product-innovations-yubikey-5-7-and-yubico-authenticator-7/) new firmware with expanded key type support. They've also [added support to their tooling](https://github.com/Yubico/yubikey-manager/blob/main/yubikit/piv.py#L89).

Would you accept PRs to add these new non-standard key types once hardware is available to test them?

maraino:
This commit supports the new algorithms supported on YubiKeys with a 5.7.x firmware. It adds support for RSA-3072, RSA-4096, Ed25519, and X25519.

Generating X25519 keys is only supported with Go 1.20+, which adds support for the `crypto/ecdh` package.

Note that this commit removes the support for Ed25519 on SoloKeys if the algorithm identifier is set to 0x22.

Fixes #143


v2/piv/key_go120.go
```diff
@@ -0,0 +1,116 @@
...
+//
+// Peer's public key must use the same algorithm as the key in this slot, or an
+// error will be returned.
+func (k *X25519PrivateKey) SharedKey(peer *ecdh.PublicKey) ([]byte, error) {
+	return k.auth.do(k.yk, k.pp, func(tx *scTx) ([]byte, error) {
+		return ykECDHX25519(tx, k.slot, k.pub, peer)
+	})
+}
+
+func (yk *YubiKey) privateKey(slot Slot, public crypto.PublicKey, auth KeyAuth, pp PINPolicy) (crypto.PrivateKey, error) {
```
hslatman: Maybe this function name can be more descriptive. Something like `tryX25519PrivateKey`? Currently it reads like a generic function that handles more cases. Also applies to the other function.

v2/piv/key_go120.go
```diff
@@ -0,0 +1,116 @@
...
+		if priv.Curve() != ecdh.X25519() {
+			return nil, 0, 0, errors.New("unsupported private key type")
+		}
+		// seed
+		params := make([][]byte, 0)
+		params = append(params, priv.Bytes())
+		return params, 0x08, 32, nil
+	default:
+		return nil, 0, 0, errors.New("unsupported private key type")
+	}
```
hslatman: Could include the actual type for context?

v2/piv/key_go120.go
```diff
@@ -0,0 +1,116 @@
...
+		return nil, fmt.Errorf("unsupported public key type: %T", public)
+	}
+}
+
+func (yk *YubiKey) setPrivateKeyInsecure(private crypto.PrivateKey) ([][]byte, byte, int, error) {
+	switch priv := private.(type) {
+	case *ecdh.PrivateKey:
+		if priv.Curve() != ecdh.X25519() {
+			return nil, 0, 0, errors.New("unsupported private key type")
+		}
```
hslatman: Should this say something like `"unsupported curve ecdh curve: %v"`? And if so, could include the curve name.

v2/piv/key_go120_test.go
```diff
@@ -0,0 +1,120 @@
...
+	"crypto/rand"
+	"errors"
+	"reflect"
+	"testing"
+)
+
+func TestYubiKeyX25519ImportKey(t *testing.T) {
+	importKey, err := ecdh.X25519().GenerateKey(rand.Reader)
+	if err != nil {
+		t.Fatalf("error geneating X25519 key: %v", err)
```
hslatman:
```suggestion
		t.Fatalf("error generating X25519 key: %v", err)
```

v2/piv/key_legacy.go
```diff
@@ -0,0 +1,36 @@
...
+	"errors"
+	"fmt"
+)
+
+func (yk *YubiKey) privateKey(slot Slot, public crypto.PublicKey, auth KeyAuth, pp PINPolicy) (crypto.PrivateKey, error) {
+	return nil, fmt.Errorf("unsupported public key type: %T", public)
+}
+
+func (yk *YubiKey) setPrivateKeyInsecure(private crypto.PrivateKey) ([][]byte, byte, int, error) {
+	return nil, 0, 0, errors.New("unsupported private key type")
```
hslatman: Could include the type as additional context?

v2/piv/key_go120.go
maraino: Fixed with 2837a99

v2/piv/key_go120.go
maraino: Fixed with 2837a99

v2/piv/key_go120.go
maraino: Fixed with 2837a99

v2/piv/key_go120_test.go
maraino: Fixed with 2837a99

v2/piv/key_legacy.go
maraino: Fixed with 2837a99

hslatman (approved):

v2/piv/key_legacy.go
ericchiang, line 1: https://go.dev/doc/devel/release#policy

Versions of Go earlier than 1.20 are no longer supported. Let's not bother with this?

v2/piv/key_go120.go
```diff
@@ -0,0 +1,115 @@
...
+
+package piv
+
+import (
+	"crypto"
+	"crypto/ecdh"
+	"fmt"
+)
+
+type X25519PrivateKey struct {
```
ericchiang: Please add a comment to this exported type similar to ECDSAPrivateKey, explaining that users are expected to type case private keys to this type. 

https://pkg.go.dev/github.com/go-piv/piv-go/v2/piv#ECDSAPrivateKey

v2/piv/key_go120.go
```diff
@@ -0,0 +1,115 @@
...
+//
+// Peer's public key must use the same algorithm as the key in this slot, or an
+// error will be returned.
+func (k *X25519PrivateKey) SharedKey(peer *ecdh.PublicKey) ([]byte, error) {
+	return k.auth.do(k.yk, k.pp, func(tx *scTx) ([]byte, error) {
+		return ykECDHX25519(tx, k.slot, k.pub, peer)
+	})
+}
+
+func (yk *YubiKey) tryX25519PrivateKey(slot Slot, public crypto.PublicKey, auth KeyAuth, pp PINPolicy) (crypto.PrivateKey, error) {
```
ericchiang: Go generally doesn't use "try" in the method name

```
newX25519PrivateKey()
```

v2/piv/key_go120.go
```diff
@@ -0,0 +1,115 @@
...
+		if crv := pub.Curve(); crv != ecdh.X25519() {
+			return nil, fmt.Errorf("unsupported ecdh curve: %v", crv)
+		}
+		return &X25519PrivateKey{yk, slot, pub, auth, pp}, nil
+	default:
+		return nil, fmt.Errorf("unsupported public key type: %T", public)
+	}
+}
+
+func (yk *YubiKey) tryX22519PrivateKeyInsecure(private crypto.PrivateKey) ([][]byte, byte, int, error) {
```
ericchiang: "setX22519PrivateKeyInsecure"

v2/piv/key_legacy_test.go
ericchiang, line 1: ditto: let's remove this and drop support for earlier versions of Go

ericchiang (approved):
This looks great! A couple of comments, but otherwise lgtm

AGWA:
Instead of adding a new exported type `X25519PrivateKey`, I propose the following:

1. Define the following interface (which is also satisfied by `*ecdh.PrivateKey`): 
```go
type ECDHer interface {
    ECDH(*ecdh.PublicKey) ([]byte, error)
}
```
2. Rename `SharedKey` to `ECDH`.
3. Add an `ECDH` function to `ECDSAPrivateKey` that satisfies this interface.
4. Deprecate `ECDSAPrivateKey.SharedKey`
5. Update the documentation for `YubiKey.PrivateKey` to say "The returned key implements crypto.Signer, crypto.Decrypter, and/or ECDHer" depending on the key type. 

`*ecdh.PrivateKey`s and hardware-backed ECDH keys would implement the same interface, providing the same flexibility that signing and decrypting keys offer via `crypto.Signer` and `crypto.Decrypter`.

maraino:
@ericchiang what do you think about @AGWA proposal? My initial implementation was like that, but I noticed that ECDSA was using SharedKey, so I changed it.

@AGWA I'm open to renaming SharedKey to ECDH if @ericchiang is ok with it. But it isn't necessary to add the interface here. The software using this package can create its own.

v2/piv/key_go120.go
maraino: If I remove support for versions before 1.20, instead of renaming it, I will add the code to PrivateKey.

v2/piv/key_go120.go
maraino: Same here. I will add the code to `SetPrivateKeyInsecure`

ericchiang:
I agree that we shouldn't create new interfaces here. At the point where you use the interface, you have to be aware of piv-go anyway. If this is going to be a common definition, it probably belongs in https://pkg.go.dev/crypto, similar to crypto.Signer. 

I'd prefer keeping "SharedKey" to be consistent within this package with ECDSAPrivateKey. Not thrilled with deprecating things for deprecation sake

AGWA:
I agree that the interface doesn't need to be defined here, but I'm still hoping we can call the function ECDH for consistency with `ecdh.PrivateKey`.

I suggested deprecating `ECDSAPrivateKey.SharedKey` because doing ECDH with `crypto/elliptic` and `crypto/ecdsa` has been deprecated in the standard library in favor of `crypto/ecdh`. The new `ECDSAPrivateKey.ECDH` method would take a `*ecdh.PublicKey` instead of a `*ecdsa.PublicKey`.  I've submitted a PR #158 for discussion.  I'm not trying to create pointless churn here; I'm working on a project where I would like to use hardware and software ECDH keys interchangeably and having a common, modern interface for ECDH would be extremely convenient compared to `ecdh.PrivateKey`, `ECDSAPrivateKey` and `X25519PrivateKey` all having a different interface.

ericchiang:
Thanks https://github.com/go-piv/piv-go/pull/158 is helpful! I didn't realize the proposal is to change the signature, not just the name. I'm fine with supporting the newer `crypto/ecdh` package keys through a "ECDH" method.

The parts of `crypto/elliptic` that were deprecated were those used to do ecdh, getting `math/big` out of the crypto paths for the standard library. piv-go doesn't use it this way (DH is done on the hardware). I don't see the need to deprecate the SharedKey method.

I'll merge #158. @maraino sorry for the flip flop, but the name "ECDH" does seem reasonable if we're accepting  `*ecdh.PublicKey`

v2/piv/key_legacy.go
maraino: Fixed with cfec690

v2/piv/key_go120.go
maraino: Fixed with cfec690

v2/piv/key_go120.go
maraino: Fixed with cfec690

v2/piv/key_go120.go
maraino: Fixed with cfec690

maraino:
@ericchiang, I've fixed the issues with my latest commit cfec690. I didn't implement the `func (*ECDSAPrivateKey) ECDH(*ecdh.PublicKey)`. I might do this in a new commit to fix the new issue created by @AGWA.

AGWA:
@maraino no need to implement `func (*ECDSAPrivateKey) ECDH(*ecdh.PublicKey)` - my PR for that has already been merged.

ericchiang (approved):

ericchiang:
Thanks everyone for the feedback here! I've gone ahead and tagged v2.2.0 including these changes https://github.com/go-piv/piv-go/releases/tag/v2.2.0

PR #158, opened 2024-10-02
Add ECDSAPrivateKey.ECDH which takes a *ecdh.PublicKey

AGWA:
The current `ECDSAPrivateKey.SharedKey` method takes a `*ecdsa.PublicKey`. However, using crypto/ecdsa and crypto/elliptic for ECDH has been [deprecated in the standard library in favor of crypto/ecdh](https://github.com/golang/go/issues/52221).

This commit adds a new ECDH method to `ECDSAPrivateKey` which takes a `*ecdh.PublicKey`.  This method has the same signature as `ecdh.PrivateKey.ECDH`, meaning the following interface can be be used to do ECDH with both standard library private keys and piv-go keys, providing the same flexibility as `crypto.Signer` and `crypto.Decrypter`:

```go
interface {
	ECDH(*ecdh.PublicKey) ([]byte, error)
}
```

`ECDSAPrivateKey.SharedKey` has been re-implemented as a small wrapper around `ECDSAPrivateKey.ECDH`.  It has not been marked as deprecated, but it could be.

### DIFF
--- a/v2/piv/key.go
+++ b/v2/piv/key.go
@@ -17,6 +17,7 @@ package piv
 import (
 	"bytes"
 	"crypto"
+	"crypto/ecdh"
 	"crypto/ecdsa"
 	"crypto/ed25519"
 	"crypto/elliptic"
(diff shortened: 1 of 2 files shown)


ericchiang (approved):
thanks!

maraino:
It's quite easy to convert on both directions `crypto/ecdh` key to `crypto/elliptic`. See, for example, this, https://github.com/smallstep/crypto/blob/47febcd8eeae3b35809606fe1f148babe38dc4e0/kms/mackms/mackms.go#L1184-L1208

We can use a method like this to implement `ECDH(*ecdh.PublicKey)`, converting the public key to ecdsa.PublicKey and use SharedKey to perform ECDH.

Broader support in PrivateKey and SetPrivateKeyInsecure is possible too, for example, if we pass as ecdh.P256 key to PrivateKey we can return a type that implements ECDH and returns the ecdh.PublicKey in the `Public() crypto.PublicKey` method. We can also implement `Curve() ecdh.Curve` or `PublicKey() *ecdh.PublicKey` but methods like `Equal(x crypto.PrivateKey) bool` will be harder to support, we can compare the public keys and assume there are no collisions :)

PR #159, opened 2024-10-04
Ed25519 improvements

AGWA:
This PR includes a few minor cleanups, and most importantly, returns an error if Ed25519ctx or Ed25519ph is requested. Previously, an invalid signature was being returned. As far as I can tell, the YubiKey does not support these Ed25519 variants.


ericchiang (approved):
thanks!

PR #160, opened 2024-11-07
Option to make yubikey connection shared

sgrb:
Add `Shared` field to `Client` struct, which switches access mode to `SCARD_SHARE_SHARED`.

It is required if the Yubikey is used by multiple applications (for example, by VPN client and yubikey-agent). Currently used `SCARD_SHARE_EXCLUSIVE` mode prevents opening yubikey if any other app (e.g, openvpn) is curently using the key.

### DIFF
--- a/piv/pcsc_test.go
+++ b/piv/pcsc_test.go
@@ -63,7 +63,7 @@ func runHandleTest(t *testing.T, f func(t *testing.T, h *scHandle)) {
 		if reader == "" {
 			t.Skip("could not find yubikey, skipping testing")
 		}
-		h, err := c.Connect(reader)
+		h, err := c.Connect(reader, false)
 		if err != nil {
 			t.Fatalf("connecting to %s: %v", reader, err)
 		}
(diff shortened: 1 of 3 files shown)


ericchiang:
Hey @sgrb, thanks for the PR!

One of the complications in shared mode is that we can't assume that another client hasn't changed the application. I believe that means we'd want to call ykSetApplication every time this library makes a call:

https://github.com/go-piv/piv-go/pull/108/files#diff-df0547c1ba82a52ddc813111d1b6343d3a9715ec46154460b1d5685decc0065eR120

So I think this requires more significant updates to the codebase?

On a related note, I'm in the processes of getting a newer yubikey to test some of more recent firmwares and might be able to take up #108 again.

sgrb:
> One of the complications in shared mode is that we can't assume that another client hasn't changed the application. I believe that means we'd want to call ykSetApplication every time this library makes a call:

Maybe you're right, I'm not an expert in Yubikey API. But at least I use this configuration (shared mode for yubikey-agent and openvpn simultaneously) without problems every day for more than a year. Perhaps it works for me because I don't use other applications - only PIV and OTP (which actually emulates input device and doesn't require application change, I suppose).

Your PR of course looks more elaborate, so I'll close this one.

ericchiang:
Sounds good, thanks for the note that this mode has been working for you!

PR #161, opened 2024-11-21
feat: add GenerateKeyRaw method in order to save the raw bytes for encrypt without the piv device

izouxv:
feat: add GenerateKeyRaw method in order to save the raw bytes for encrypt, and set DecodePublic to public for parse the raw bytes

### DIFF
--- a/piv/key.go
+++ b/piv/key.go
@@ -641,7 +641,7 @@ func (ki *KeyInfo) unmarshal(b []byte) error {
 				return errors.New("unknown origin in response")
 			}
 		case 4:
-			ki.PublicKey, err = decodePublic(v.Bytes, ki.Algorithm)
+			ki.PublicKey, err = DecodePublic(v.Bytes, ki.Algorithm)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


ericchiang:
Hey, thanks for the PR! Can you elaborate on why you need this? Is it for a public key type this package doesn't support, or some other reason?

izouxv:
> Hey, thanks for the PR! Can you elaborate on why you need this? Is it for a public key type this package doesn't support, or some other reason?

 save the pub key for encrypt something without piv device.

ericchiang:
@izouxv you should already be able to serialize the key with encodings like https://pkg.go.dev/crypto/x509#MarshalPKIXPublicKey

That works fine without this package exposing the raw public key or parsing logic

izouxv:
> @izouxv you should already be able to serialize the key with encodings like https://pkg.go.dev/crypto/x509#MarshalPKIXPublicKey
> 
> That works fine without this package exposing the raw public key or parsing logic

thank you

PR #164, opened 2024-12-23
export KeyRSA and KeyEd25519

tomis007:
I think `keyRSA` and `keyEd25519` are supposed to be Public types but aren't exported because of their lowercase name. Here's a PR to fix it. I had to change it to get `piv-go` working with my yubikey with RSA2048 keys.

### DIFF
--- a/v2/piv/key.go
+++ b/v2/piv/key.go
@@ -1295,9 +1295,9 @@ func (k *keyEd25519) Public() crypto.PublicKey {
 	return k.pub
 }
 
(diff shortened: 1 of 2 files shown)


ericchiang:
This is intentional. crypto.PrivateKey is supposed to be type cased to crypto.Signer and crypto.Decrypter as those methods are needed. Since that's supported, the implementations aren't exported

https://pkg.go.dev/crypto#PrivateKey

tomis007:
Thank you! I did that and it worked, I added a pr for the readme https://github.com/go-piv/piv-go/pull/165 if you think that's useful for others

PR #166, opened 2025-01-20
feat: allow to read and write raw PIV objects

impact-merlinzerbe:
the YubiKey allows to store raw data in PIV DataTags (see https://docs.yubico.com/yesdk/users-manual/application-piv/piv-objects.html). example use cases are storage of gpg public key counterparts of gpg private keys and encrypted storage of other key data (e.g. wireguard keys). this feature is also available in the official ykman app from Yubico (ykman piv objects import/export).

### DIFF
--- a/v2/piv/key.go
+++ b/v2/piv/key.go
@@ -688,6 +688,35 @@ func (yk *YubiKey) KeyInfo(slot Slot) (KeyInfo, error) {
 	return ki, nil
 }
 
+// Object returns the raw object stored in a given slot.
+func (yk *YubiKey) Object(slot Slot) ([]byte, error) {
+	cmd := apdu{
+		instruction: insGetData,
+		param1:      0x3f,
+		param2:      0xff,
+		data: []byte{
+			0x5c, // Tag list
+			0x03, // Length of tag
+			byte(slot.Object >> 16),
+			byte(slot.Object >> 8),
+			byte(slot.Object),
+		},
+	}
+
+	resp, err := yk.tx.Transmit(cmd)
+	if err != nil {
+		return nil, fmt.Errorf("command failed: %w", err)
+	}
+
+	// https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-73-4.pdf#page=85
+	obj, _, err := unmarshalASN1(resp, 1, 0x13) // tag 0x53
+	if err != nil {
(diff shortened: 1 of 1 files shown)


ericchiang:
Thanks for the PR!

Couple of questions:
- What other systems might validate that the object is encoded correctly? Does the yubikey firmware do this, or is it up to other applications?
- Can there be multiple objects per slot, or only one?
- Should the API differentiate tag values from the object? E.g. `SetObject(key []byte, slot Slot, tag uint64, obj []byte)`

EDIT: The CI issues you're seeing should be fixed by #167

impact-merlinzerbe:
Thanks for the quick response.

> * What other systems might validate that the object is encoded correctly? Does the yubikey firmware do this, or is it up to other applications?

AFAIK the only the the yubikey validates is the initial tag and the length (https://docs.yubico.com/yesdk/users-manual/application-piv/piv-objects.html#the-data)

```
53 length
   something
```

> * Can there be multiple objects per slot, or only one?

I think there can only be a single object per slot, but there are over 6 million possible slots. The total slot storage space is limited to 51000 bytes (https://docs.yubico.com/yesdk/users-manual/application-piv/piv-objects.html#table-1c-undefined-datatags):

> It is possible for an application to store whatever information it wants on a YubiKey under an undefined DataTag. However, there are space limitations. It is possible to store at most approximately 3,052 bytes under any single undefined DataTag, and the total space on a YubiKey for all storage is about 51,000 bytes.

> * Should the API differentiate tag values from the object? E.g. `SetObject(key []byte, slot Slot, tag uint64, obj []byte)`

I do not know. I currently use this API to store raw []byte (encrypted blobs). Could you elaborate what you mean by differentiating?

impact-merlinzerbe:
@ericchiang is there anything I can do to move this forward?

bherw:
I'm interested in using these APIs as well. Is there anything I can do to help move this forward?

> Should the API differentiate tag values from the object? E.g. `SetObject(key []byte, slot Slot, tag uint64, obj []byte)`

I don't think it makes sense to have both slot and tag parameters. The current API proposed by @impact-merlinzerbe is specifying the 3-byte tag using the slot parameter. The specification seems to differentiate `tag` from `slot`, and `tag` is 32-bits, so to match the spec, `WriteObject(key []byte, tag uint32, obj []byte)`.

PR #171, opened 2025-04-15
Private Ed25519 and RSA keys as public crypt.Signer 

cviecco:
As part of another project I wanted crypto.Signers to be provided by the piv-go package. Today only the ECDSA private key i(ECDSAPrivateKey) is public. Thus I am assuming the authors are OK with having other public private keys.

This patch is the minimum code to make this happen by creating an public alias to the current internal struct and ensuring that these internal strut are atually crypto.Signers.

I think this is cleaner than renaming the internal structs. But If you prefer that I can generate such patch.

### DIFF
--- a/v2/piv/key.go
+++ b/v2/piv/key.go
@@ -1291,9 +1291,13 @@ type keyEd25519 struct {
 	pp   PINPolicy
 }
 
+type Ed25519Key = keyEd25519
+
 func (k *keyEd25519) Public() crypto.PublicKey {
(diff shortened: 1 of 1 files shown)


ericchiang:
See previous discussion:

https://github.com/go-piv/piv-go/pull/164#issuecomment-2558815207

This is intentional. Users should type cast to crypto.Signer and crypto.Decrypter, not depend on concrete types

PR #181, opened 2025-07-12
Connect with `SCARD_SHARE_SHARED`

### LINKED ISSUE #116: Error: Unable to open the Yubikey card: connecting to smart card: the smart card cannot be accessed because of other connections outstanding
Hello

I'm able to build a program with this library that allows me to successfully interact with my YubiKey. However under some conditions, e.g. after some idle period, I'm getting the aforementioned erorr:
```
Error: Unable to open the Yubikey card: connecting to smart card: the smart card cannot be accessed because of other connections outstanding
```

I'm not sure if there is anything wrong with piv-go. But I'm quite puzzled since I make thorough use of the `Close()` method and additionally I don't get this error with other tools e.g. libykcs11.so.

DavidBuchanan314:
Fixes #116

However, there may be other consequences to this change. I don't know what they might be. Hopefully someone with deeper knowledge of the relevant APIs can weigh in.

All I know is that ykpiv also uses `SCARD_SHARE_SHARED`: https://github.com/Yubico/yubico-piv-tool/blob/73db815e8927028c51ba71771a6737efaa238a62/lib/ykpiv.c#L955

### DIFF
--- a/v2/piv/pcsc_unix.go
+++ b/v2/piv/pcsc_unix.go
@@ -98,7 +98,7 @@ func (c *scContext) Connect(reader string) (*scHandle, error) {
 		activeProtocol C.DWORD
 	)
 	rc := C.SCardConnect(c.ctx, C.CString(reader),
-		C.SCARD_SHARE_EXCLUSIVE, C.SCARD_PROTOCOL_T1,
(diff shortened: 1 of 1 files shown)


DavidBuchanan314:
Ah, just saw https://github.com/go-piv/piv-go/pull/108 - looks like a much more thought-out approach

PR #185, opened 2026-02-02
Add PIV biometric match policies and VerifyUV

jmecom:


v2/piv/key.go
```diff
@@ -485,6 +489,11 @@ const (
 	PINPolicyNever PINPolicy = iota + 1
 	PINPolicyOnce
 	PINPolicyAlways
+	// PINPolicyMatchOnce and PINPolicyMatchAlways require biometric user
+	// verification (YubiKey Bio) via VerifyUV. They are supported by
+	// YubiKey Bio Multi-protocol devices.
+	PINPolicyMatchOnce
```
ericchiang, line 502: Are these the official names of the policies? It seems odd that they'd be so similar to "PINPolicyOnce"

jmecom: Good point; that name is pretty bad. I put the same prefix, and got the "match once / always" part from here: https://docs.yubico.com/yesdk/users-manual/application-piv/pin-touch-policies.html

I'll change to `TouchPolicy*`. Also realized I didn't add unit tests, so will add that too.

edit: Actually, `TouchPolicy` is misleading too, because touch != biometric match.

jmecom: I'm also unsure about the match-once semantics for touch, I'll request re-review once I've tested this more.

ericchiang: Should these be called something like: BiometricPolicyMatchOnce, BiometricPolicyMatchAlways?

I assume these are exclusive to other pin policies?

v2/piv/piv.go
```diff
@@ -243,6 +244,14 @@ func (yk *YubiKey) VerifyPIN(pin string) error {
 	return ykLogin(yk.tx, pin)
 }
 
+// VerifyBiometrics performs biometric user verification (YubiKey Bio) for PIV.
```
ericchiang, line 247: This library intentionally doesn't require the user to reason about when they have to authenticate. It just authenticates using the correct strategy for each call.

Can this be done transparently? E.g. by calling ykVerifyUV at the appropriate time rather than forcing the user to determine when they need to?

v2/piv/bio_integration_test.go
```diff
@@ -0,0 +1,287 @@
...
+		return
+	}
+}
+
+func signWithoutBioExpected(t *testing.T, signer crypto.Signer, message []byte) {
+	out, closeTTY := openTTY()
+	if closeTTY != nil {
+		defer closeTTY()
+	}
+	fmt.Fprintf(out, "Signing message: %s\n", message)
```
ericchiang: Please take these debug statements out. Tests generally shouldn't do this

v2/piv/bio_integration_test.go
```diff
@@ -0,0 +1,287 @@
...
+	digest := sha256.Sum256(message)
+	sig, err := signer.Sign(rand.Reader, digest[:], crypto.SHA256)
+	if err != nil {
+		if isUnsupportedMatchPolicy(err) {
+			t.Skipf("match policy not supported: %v", err)
+		}
+		t.Fatalf("expected signing to succeed without biometric verification: %v", err)
+	}
+	fmt.Fprintf(out, "Signed: %x\n", sig)
```
ericchiang: ditto: please remove this. if it's important, very the signature in the test

v2/piv/bio_integration_test.go
```diff
@@ -0,0 +1,287 @@
...
+	if closeTTY != nil {
+		defer closeTTY()
+	}
+	fmt.Fprintln(out, msg)
+}
+
+func openTTY() (*os.File, func()) {
+	tty, err := os.OpenFile("/dev/tty", os.O_RDWR, 0)
+	if err != nil {
+		return os.Stderr, nil
```
ericchiang: I would error here and fail the test if you can't open the tty

v2/piv/bio_integration_test.go
```diff
@@ -0,0 +1,287 @@
...
+	}
+	fmt.Fprintf(out, "Signed: %x\n", sig)
+}
+
+func signAndPrint(t *testing.T, signer crypto.Signer, message []byte) {
+	out, closeTTY := openTTY()
+	if closeTTY != nil {
+		defer closeTTY()
+	}
+	fmt.Fprintf(out, "Signing message: %s\n", message)
```
ericchiang: ditto: please remove debug statements

v2/piv/bio_integration_test.go
```diff
@@ -0,0 +1,287 @@
...
+	digest := sha256.Sum256(message)
+	sig, err := signer.Sign(rand.Reader, digest[:], crypto.SHA256)
+	if err != nil {
+		if isUnsupportedMatchPolicy(err) {
+			t.Skipf("match policy not supported: %v", err)
+		}
+		t.Fatalf("signing failed: %v", err)
+	}
+	fmt.Fprintf(out, "Signed: %x\n", sig)
```
ericchiang: same here

v2/piv/bio_integration_test.go
```diff
@@ -0,0 +1,287 @@
...
+}
+
+func confirmDestructiveSlot9a(t *testing.T) {
```
ericchiang: This is already implied by ` -wipe-yubikey`, right? Why do you need another prompt?

v2/piv/key.go
jmecom: I'm cool with changing the name. I went back-and-forth here because the `PINPolicy` prefix does match `yubico-piv-tool`'s naming conventions. See here: https://github.com/Yubico/yubico-piv-tool/blob/master/lib/ykpiv.h#L770-L771

But of course, that naming is also because it's C and therefore doesn't have real namespaces. So, that is not the strongest justification for this codebase to match its naming.

I lean toward keeping the name as it is, but I think it's your call.

v2/piv/bio_integration_test.go
jmecom: I was intentionally redundant to be extra cautious, since it's a destructive operation, but I'll remove it (along with the prints etc).

jmecom:
@ericchiang Just pushed a commit which should address your API concerns. A few notes:

* The YK hardware doesn't seem to have a way to query UV (user verification) status. So the best option for "match once" keys is to try the operation, check if it fails, and then try again.
* I left the prompt to touch the fingerprint sensor in the tests, but removed the other logs. I think that prompt is kind of necessary, otherwise the test is confusing, as it's easy to not realize you need to perform UV -- and then the test fails.

v2/piv/key.go
ericchiang: If this matches ykpiv, I'm happy to use that. Thanks for the link!

ericchiang:
> So the best option for "match once" keys is to try the operation, check if it fails, and then try again.

Lol, of course... Thanks for looking into this! If this works, I'm happy to submit as is.

> I left the prompt to touch the fingerprint sensor in the tests, but removed the other logs. I think that prompt is kind of necessary, otherwise the test is confusing, as it's easy to not realize you need to perform UV -- and then the test fails.

NP! The tests for this package were always pretty manual. I just wanted to make sure that we're not logging too much stuff

Thanks for your contribution here. If you can squash your commits, I can merge and cut a release later today

ericchiang:
(apologies, I didn't get a notification for the force push. merging now!)

ericchiang:
https://github.com/go-piv/piv-go/releases/tag/v2.5.0 tagged

PR #187, opened 2026-04-15
Obtain the form factor of the Yubikey

smlx:
Use the management applet to obtain the form factor of the yubikey.

### DIFF
--- a/v2/piv/key.go
+++ b/v2/piv/key.go
@@ -34,6 +34,7 @@ import (
 	"strings"
 
 	rsafork "github.com/go-piv/piv-go/v2/third_party/rsa"
+	"golang.org/x/crypto/cryptobyte"
 
 	_ "embed"
 )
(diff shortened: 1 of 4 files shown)


v2/piv/key.go
```diff
@@ -765,6 +774,125 @@ func (yk *YubiKey) Certificate(slot Slot) (*x509.Certificate, error) {
 	return cert, nil
 }
 
+// FormFactor returns the physical form factor of the YubiKey and a
+// human-readable description.
+func (yk *YubiKey) FormFactor() (Formfactor, string, error) {
```
ericchiang: Formfactor already has a String() parameter that returns the human readable name, so the second argument shouldn't be needed here

https://github.com/go-piv/piv-go/blob/65bdacedce76cc718f3968e5b6f805ddf19b720d/v2/piv/key.go#L104-L109

```
// FormFactor returns the physical form factor of the YubiKey.
func (yk *YubiKey) FormFactor() (Formfactor, error)
```

ericchiang (approved):
Just one nit, thanks!

ericchiang:
You should be able to rebase on top of https://github.com/go-piv/piv-go/pull/188 for the CI issues

v2/piv/key.go
smlx: Oh wow, not sure how I missed that. Fixed!

ericchiang:
https://github.com/go-piv/piv-go/releases/tag/v2.6.0

Thanks!

